PR #6, opened 2021-07-12
Moved PR for references impl from core to event core.

douradofrazer:
Resolves:
https://dcsa.atlassian.net/browse/DDT-25


douradofrazer:
@nt-as do we need to bump up the version? its not a breaking/major change so I didn't bump it up for now.

nt-as (approved):

PR #7, opened 2021-07-14
Sync changes for IM 3.2

douradofrazer:
Resolves:
https://dcsa.atlassian.net/browse/DDT-38


nt-as (approved):

src/main/java/org/dcsa/core/events/model/ShipmentEvent.java
```diff
@@ -20,8 +20,13 @@
     @Column("document_type_code")
     private DocumentTypeCode documentTypeCode;
 
-    @Column("shipment_id")
-    private UUID shipmentID;
+    @Column("document_id")
+    @Size(max = 100)
+    private UUID documentId;
```
nt-as, line 25:
```suggestion
    private String documentID;
```

nt-as (changes requested):
Wrong type and name for documentID attribute

nt-as (approved):

PR #10, opened 2021-07-15
Map in References, DocumentReferences and TransportCallTO on EquipmentEvent

nt-as:


nt-as:
The transport call object is still incomplete. That is a known issue.

douradofrazer (approved):

PR #15, opened 2021-07-19
Updated shipment event with document references and references

douradofrazer:
Resolves:
https://dcsa.atlassian.net/browse/DDT-28

### DIFF
--- a/src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
+++ b/src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
@@ -42,4 +42,10 @@ public interface ReferenceRepository extends ExtendedRepository<Reference, UUID>
           + "JOIN transport_document td2 ON  td2.shipping_instruction_id = si.id "
           + "WHERE td2.transport_document_reference = :transportDocumentReference ) ")
   Flux<Reference> findByTransportDocumentReference(String transportDocumentReference);
+
+  @Query(
+      "SELECT \"references\".* FROM \"references\" "
+          + "JOIN shipment s ON shipment_id = s.id "
+          + "WHERE s.carrier_booking_reference = :carrierBookingReference")
+  Flux<Reference> findByCarrierBookingReference(String carrierBookingReference);
 }
--- a/src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
+++ b/src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
@@ -45,7 +45,7 @@ public interface ReferenceRepository extends ExtendedRepository<Reference, UUID>
 
   @Query(
       "SELECT \"references\".* FROM \"references\" "
-          + "JOIN shipment s ON shipment_id = s.id "
+          + "LEFT JOIN shipment s ON shipment_id = s.id "
           + "WHERE s.carrier_booking_reference = :carrierBookingReference "
           + "OR shipping_instruction_id IN ( SELECT si.id from shipping_instruction si "
           + "JOIN cargo_item ci ON ci.shipping_instruction_id = si.id "
--- a/src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
+++ b/src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
@@ -46,6 +46,10 @@ public interface ReferenceRepository extends ExtendedRepository<Reference, UUID>
   @Query(
       "SELECT \"references\".* FROM \"references\" "
           + "JOIN shipment s ON shipment_id = s.id "
-          + "WHERE s.carrier_booking_reference = :carrierBookingReference")
+          + "WHERE s.carrier_booking_reference = :carrierBookingReference "
+          + "OR shipping_instruction_id IN ( SELECT si.id from shipping_instruction si "
+          + "JOIN cargo_item ci ON ci.shipping_instruction_id = si.id "
+          + "JOIN shipment s ON ci.shipment_id = s.id "
+          + "WHERE s.carrier_booking_reference = :carrierBookingReference)")
   Flux<Reference> findByCarrierBookingReference(String carrierBookingReference);
 }
--- a/src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
+++ b/src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
@@ -35,12 +35,13 @@ public interface ShipmentEventRepository extends ExtendedRepository<ShipmentEven
       "SELECT DISTINCT td.transport_document_reference FROM transport_document td "
           + "JOIN shipping_instruction si ON td.shipping_instruction_id  = si .id "
           + "JOIN cargo_item ci ON si.id = ci.shipping_instruction_id "
-          + "WHERE ci.shipment_id = :shipmentId")
-  Flux<String> findTransportDocumentRefsByShipmentId(UUID shipmentId);
+          + "WHERE ci.shipping_instruction_id = :shippingInstructionID")
+  Flux<String> findTransportDocumentRefsByShippingInstructionID(String shippingInstructionID);
 
   @Query(
       "SELECT DISTINCT b.carrier_booking_reference FROM booking b "
           + "JOIN shipment s ON b.carrier_booking_reference = s.carrier_booking_reference "
-          + "WHERE s.id = :shipmentId")
-  Flux<String> findCarrierBookingRefsByShipmentId(UUID shipmentId);
+          + "JOIN cargo_item ci ON s.id = ci.shipment_id "
+          + "WHERE ci.shipping_instruction_id = :shippingInstructionID")
+  Flux<String> findCarrierBookingRefsByShippingInstructionID(String shippingInstructionID);
 }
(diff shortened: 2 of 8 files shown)


src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
```diff
@@ -42,4 +42,10 @@
           + "JOIN transport_document td2 ON  td2.shipping_instruction_id = si.id "
           + "WHERE td2.transport_document_reference = :transportDocumentReference ) ")
   Flux<Reference> findByTransportDocumentReference(String transportDocumentReference);
+
+  @Query(
+      "SELECT \"references\".* FROM \"references\" "
+          + "JOIN shipment s ON shipment_id = s.id "
+          + "WHERE s.carrier_booking_reference = :carrierBookingReference")
+  Flux<Reference> findByCarrierBookingReference(String carrierBookingReference);
```
nt-as, line 50: This is too simple - it should also find references on transport document related to that shipment.  This query should look similar to the one for `findByShipmentID` (plus JOIN to `shipment` so you can use `carrier_booking_reference` in the `WHERE` clause)

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
```diff
@@ -13,4 +13,34 @@
...
+          + "JOIN transport_document td ON si.id = td.shipping_instruction_id "
+          + "WHERE td.transport_document_reference = :transportDocumentRef")
+  Flux<String> findCarrierBookingRefsByTransportDocumentRef(String transportDocumentRef);
+
+  @Query(
+      "SELECT DISTINCT td.transport_document_reference FROM transport_document td "
+          + "JOIN shipping_instruction si ON td.shipping_instruction_id  = si .id "
+          + "JOIN cargo_item ci ON si.id = ci.shipping_instruction_id "
+          + "WHERE ci.shipment_id = :shipmentId")
+  Flux<String> findTransportDocumentRefsByShipmentId(UUID shipmentId);
```
nt-as, line 39:
```suggestion
          + "WHERE ci.shipment_id = :shipmentID")
  Flux<String> findTransportDocumentRefsByShipmentID(UUID shipmentID);
```
(Will need changes to the caller as well)

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
```diff
@@ -13,4 +13,34 @@
...
+
+  @Query(
+      "SELECT DISTINCT b.carrier_booking_reference FROM booking b "
+          + "JOIN shipment s ON b.carrier_booking_reference = s.carrier_booking_reference "
+          + "WHERE s.id = :shipmentId")
+  Flux<String> findCarrierBookingRefsByShipmentId(UUID shipmentId);
```
nt-as, line 45:
```suggestion
          + "WHERE s.id = :shipmentID")
  Flux<String> findCarrierBookingRefsByShipmentID(UUID shipmentID);
```

Remember to update callers as well (not covered in my review)

src/main/java/org/dcsa/core/events/service/impl/ShipmentEventServiceImpl.java
```diff
@@ -26,4 +34,78 @@ public ShipmentEventRepository getRepository() {
...
+                    transformDocRefs
+                        .apply(
+                            DocumentReferenceType.BKG,
+                            Flux.merge(
+                                shipmentEventRepository
+                                    .findTransportDocumentRefsByCarrierBookingRef(
+                                        se.getDocumentID()),
+                                Flux.just(se.getDocumentID())))
+                        .doOnNext(se::setDocumentReferences)
+                        .subscribe());
```
nt-as, line 57: `.subscribe()` cannot be used here - it will resolve "too late".

Replace the `.doOnNext` on l46 instead with a `.flatMap` and return the Mono returned by `transformDocRefs` instead (possibly chaining into a `.thenReturn` if you need to change the type).

src/main/java/org/dcsa/core/events/service/impl/ShipmentEventServiceImpl.java
```diff
@@ -26,4 +34,78 @@ public ShipmentEventRepository getRepository() {
...
+                    transformDocRefs
+                        .apply(
+                            DocumentReferenceType.TRD,
+                            Flux.merge(
+                                shipmentEventRepository
+                                    .findCarrierBookingRefsByTransportDocumentRef(
+                                        se.getDocumentID()),
+                                Flux.just(se.getDocumentID())))
+                        .doOnNext(se::setDocumentReferences)
+                        .subscribe());
```
nt-as, line 74: Ditto (subscribe)

src/main/java/org/dcsa/core/events/service/impl/ShipmentEventServiceImpl.java
```diff
@@ -26,4 +34,78 @@ public ShipmentEventRepository getRepository() {
...
+                    transformDocRefs
+                        .apply(
+                            DocumentReferenceType.SHI,
+                            Flux.merge(
+                                shipmentEventRepository.findCarrierBookingRefsByShipmentId(
+                                    se.getShipmentID()),
+                                shipmentEventRepository.findTransportDocumentRefsByShipmentId(
+                                    se.getShipmentID())))
+                        .doOnNext(se::setDocumentReferences)
+                        .subscribe());
```
nt-as, line 91: Ditto (subscribe)

src/main/java/org/dcsa/core/events/service/impl/ShipmentEventServiceImpl.java
```diff
@@ -26,4 +34,78 @@ public ShipmentEventRepository getRepository() {
...
+      default:
+        return Mono.justOrEmpty(shipmentEvent);
```
nt-as, line 93:
```suggestion
        return Mono.just(shipmentEvent);
```

src/main/java/org/dcsa/core/events/service/impl/ShipmentEventServiceImpl.java
```diff
@@ -26,4 +34,78 @@ public ShipmentEventRepository getRepository() {
...
+            .doOnNext(
+                se ->
+                    transformDocRefs
+                        .apply(
+                            DocumentReferenceType.TRD,
+                            Flux.merge(
+                                shipmentEventRepository
+                                    .findCarrierBookingRefsByTransportDocumentRef(
+                                        se.getDocumentID()),
+                                Flux.just(se.getDocumentID())))
```
nt-as, line 72: The carrier booking references should be tagged as BKG and not TRD.

src/main/java/org/dcsa/core/events/service/impl/ShipmentEventServiceImpl.java
```diff
@@ -26,4 +34,78 @@ public ShipmentEventRepository getRepository() {
...
+            .doOnNext(
+                se ->
+                    transformDocRefs
+                        .apply(
+                            DocumentReferenceType.BKG,
+                            Flux.merge(
+                                shipmentEventRepository
+                                    .findTransportDocumentRefsByCarrierBookingRef(
+                                        se.getDocumentID()),
+                                Flux.just(se.getDocumentID())))
```
nt-as, line 55: The transport document references should be tagged as TRD and not BKG.

src/main/java/org/dcsa/core/events/service/impl/ShipmentEventServiceImpl.java
```diff
@@ -26,4 +34,78 @@ public ShipmentEventRepository getRepository() {
...
+            .doOnNext(
+                se ->
+                    transformDocRefs
+                        .apply(
+                            DocumentReferenceType.SHI,
+                            Flux.merge(
+                                shipmentEventRepository.findCarrierBookingRefsByShipmentId(
+                                    se.getShipmentID()),
+                                shipmentEventRepository.findTransportDocumentRefsByShipmentId(
+                                    se.getShipmentID())))
```
nt-as, line 89: As mentioned earlier, I do not think we should use `SHI` here.  Instead BKG + TRD (in that order).

nt-as (changes requested):

src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
```diff
@@ -42,4 +42,14 @@
           + "JOIN transport_document td2 ON  td2.shipping_instruction_id = si.id "
           + "WHERE td2.transport_document_reference = :transportDocumentReference ) ")
   Flux<Reference> findByTransportDocumentReference(String transportDocumentReference);
+
+  @Query(
+      "SELECT \"references\".* FROM \"references\" "
+          + "JOIN shipment s ON shipment_id = s.id "
```
nt-as, line 48: Should probably be a `LEFT JOIN` as otherwise the reference is filtered out even if there is a valid shipping instruction ID for it (because it cannot `JOIN` with `shipment`).

(This might be a general mistake in more queries - if so, please fix them where you spot them).

Note that I am approving despite of this remark because I think we can fix this afterwards and I suspect you did not introduce this bug.

nt-as (approved):
Approved with one "should be fix but can be fixed later" remark.

src/main/java/org/dcsa/core/events/repository/ReferenceRepository.java
douradofrazer: You're right , I updated it for this query. I will update others as I spot them.

nt-as (approved):

PR #16, opened 2021-07-20
DDT-14: Create a PendingMessage model for the subscription logic

nt-as:


nt-as:
FTR: This will need to be combined with some changes in TNT to make sense

douradofrazer (approved):

PR #18, opened 2021-07-21
DDT-24 Map transport call

multimikael:
This should conclude mapping of transport call into events.

### DIFF
--- a/src/main/java/org/dcsa/core/events/model/Address.java
+++ b/src/main/java/org/dcsa/core/events/model/Address.java
@@ -16,8 +16,6 @@ import java.util.UUID;
 @NoArgsConstructor
 public class Address {
 
-    public static final Address NULL_ADDRESS = new Address();
-
     @Id
     @JsonIgnore
     private UUID id;
--- a/src/main/java/org/dcsa/core/events/model/Address.java
+++ b/src/main/java/org/dcsa/core/events/model/Address.java
@@ -16,6 +16,8 @@ import java.util.UUID;
 @NoArgsConstructor
 public class Address {
 
+    public static final Address NULL_ADDRESS = new Address();
+
     @Id
     @JsonIgnore
     private UUID id;
--- a/src/main/java/org/dcsa/core/events/model/Vessel.java
+++ b/src/main/java/org/dcsa/core/events/model/Vessel.java
@@ -1,16 +1,23 @@
 package org.dcsa.core.events.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
+import org.dcsa.core.events.model.enums.CarrierCodeListProvider;
+import org.dcsa.core.events.model.enums.FacilityCodeListProvider;
+import org.dcsa.core.events.model.transferobjects.FacilityTO;
 import org.dcsa.core.model.AuditBase;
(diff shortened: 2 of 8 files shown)


src/main/java/org/dcsa/core/events/model/transferobjects/LocationTO.java
```diff
@@ -16,12 +17,11 @@
 @NoArgsConstructor
 public class LocationTO extends AbstractLocation {
 
-    private static final Address NULL_ADDRESS = new Address();
-    private static final LocationTO NULL_LOCATION = new LocationTO();
-    private static final LocationTO NULL_LOCATION_WITH_NULL_ADDRESS = new LocationTO();
+    public static final LocationTO NULL_LOCATION = new LocationTO();
+    public static final LocationTO NULL_LOCATION_WITH_NULL_ADDRESS = new LocationTO();
```
nt-as, line 21: These should not be public as they are not immutable.

nt-as (changes requested):

src/main/java/org/dcsa/core/events/model/Address.java
```diff
@@ -16,6 +16,8 @@
 @NoArgsConstructor
 public class Address {
 
+    public static final Address NULL_ADDRESS = new Address();
```
nt-as, line 19: Should not be public as it is not immutable.

src/main/java/org/dcsa/core/events/model/Vessel.java
```diff
@@ -19,6 +26,8 @@
 @NoArgsConstructor
 public class Vessel extends AuditBase {
 
+    public static final Vessel NULL_VESSEL = new Vessel();
```
nt-as, line 29: Should not be public as it is not immutable.

src/main/java/org/dcsa/core/events/model/transferobjects/FacilityTO.java
```diff
@@ -12,9 +12,20 @@
 @EqualsAndHashCode(callSuper = true)
 public class FacilityTO extends AbstractFacility {
 
+    public static FacilityTO NULL_FACILITY = new FacilityTO();
+    public static FacilityTO NULL_FACILITY_WITH_NULL_LOCATION_WITH_NULL_ADDRESS = new FacilityTO();
```
nt-as, line 16: Should not be public as it is not immutable.

src/main/java/org/dcsa/core/events/model/Vessel.java
```diff
@@ -38,12 +42,30 @@
...
+    public void setCarrier(Carrier carrier) {
+        if (carrier != null) {
+            if (carrier.getSmdgCode() != null) {
+                vesselOperatorCarrierCode = carrier.getSmdgCode();
+                vesselOperatorCarrierCodeListProvider = CarrierCodeListProvider.SMDG;
+            } else if (carrier.getNmftaCode() != null) {
+                vesselOperatorCarrierCode = carrier.getNmftaCode();
+                vesselOperatorCarrierCodeListProvider = CarrierCodeListProvider.NMFTA;
+            }
+        }
```
nt-as, line 62: Please null `vesselOperatorCarrierCode` and `vesselOperatorCarrierCodeListProvider` if the carrier is null and add an exception if we have a carrier with no code.

nt-as (changes requested):

nt-as (approved):

nt-as (approved):

PR #23, opened 2021-07-23
DDT-32 query fields

douradofrazer:
Resolves:
https://dcsa.atlassian.net/browse/DDT-32

**Note:** scheduleID is know issue and will be fixed later.

### DIFF
--- a/src/main/java/org/dcsa/core/events/model/Event.java
+++ b/src/main/java/org/dcsa/core/events/model/Event.java
@@ -49,6 +49,7 @@ public class Event extends AuditBase {
     @Column("event_classifier_code")
     private EventClassifierCode eventClassifierCode;
 
+    @JsonIgnore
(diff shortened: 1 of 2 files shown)


src/main/java/org/dcsa/core/events/model/Event.java
```diff
@@ -52,4 +49,6 @@
     @Column("event_classifier_code")
     private EventClassifierCode eventClassifierCode;
 
+    @Column("carrier_booking_reference")
+    private String carrierBookingReference;
```
nt-as, line 53: Should be `@JsonIgnore` (the field is not in Swagger).

douradofrazer: Updated.

nt-as (approved):

PR #27, opened 2021-07-29
OperationsEvent has been moved from OVS to Event Core

kagemand-andersen:


src/main/java/org/dcsa/core/events/model/enums/PortCallServiceTypeCode.java
```diff
@@ -0,0 +1,10 @@
+package org.dcsa.core.events.model.enums;
+
+public enum PortCallServiceTypeCode {
```
HenrikHL, line 3: I know you have probably copied this code from OVS - but can you please check if we have something called PortCallServiceTypeCode in the IM?
I think it is called something different - but I might be wrong...

kagemand-andersen: As far as I can see it's called `port_call_service_type` in the IM, bottom of `03_dcsa_im_v3_0.sql`.

HenrikHL (approved):

PR #28, opened 2021-07-30
Added logic for request validation

douradofrazer:


src/main/java/org/dcsa/core/events/util/EventsRequestValidation.java
```diff
@@ -0,0 +1,152 @@
...
+            .thenRunAsync(() -> validateTransportCallID(queryParams.get("transportCallID")))
+            .thenRunAsync(() -> validateVesselIMONumber(queryParams.get("vesselIMONumber")))
+            .thenRunAsync(() -> validateCarrierVoyageNumber(queryParams.get("carrierVoyageNumber")))
+            .thenRunAsync(() -> validateCarrierServiceCode(queryParams.get("carrierServiceCode")))
+            .thenRunAsync(() -> validateLimit(queryParams.get("limit")));
+
+    return CompletableFuture.allOf(specificValidations, commonValidations);
+  }
+
+  protected void validateTNTEventType(String eventType) {
```
HenrikHL, line 45: This sounds very TnT like - shouldn't this go into TnT?

douradofrazer:
Found a better solution for this, hence this is no longer needed.

PR #34, opened 2021-08-05
DDT:198 Fix missing Mode of Transport

kagemand-andersen:


src/main/java/org/dcsa/core/events/repository/ModeOfTransportRepository.java
```diff
@@ -13,6 +13,6 @@
       "SELECT mot.* from transport t " +
```
HenrikHL, line 13: Doesn't this require a "FIRST"? If multiple `transport`s use the same transportCall - this will fail...

kagemand-andersen: Yes, I haven't pushed it yet because I ran into the Method Not Allowed 405 issue.

PR #35, opened 2021-08-05
Update ModeOfTransportRepository.java

kagemand-andersen:


src/main/java/org/dcsa/core/events/repository/ModeOfTransportRepository.java
```diff
@@ -13,6 +13,7 @@
       "SELECT mot.* from transport t " +
```
HenrikHL, line 13: you are still missing the FIRST - this will fail if transport_call_id is used in multiple transports...

kagemand-andersen: There's a `LIMIT 1` at the end of the query.

HenrikHL (approved):

PR #36, opened 2021-08-06
DDT: 208 - Add missing fields in OperationsEvent

kagemand-andersen:
Issue: https://dcsa.atlassian.net/browse/DDT-208

### DIFF
--- a/src/main/java/org/dcsa/core/events/service/impl/OperationsEventServiceImpl.java
+++ b/src/main/java/org/dcsa/core/events/service/impl/OperationsEventServiceImpl.java
@@ -4,10 +4,15 @@ import lombok.RequiredArgsConstructor;
 import org.dcsa.core.events.model.OperationsEvent;
 import org.dcsa.core.events.model.base.AbstractLocation;
 import org.dcsa.core.events.model.base.AbstractParty;
+import org.dcsa.core.events.model.base.AbstractTransportCall;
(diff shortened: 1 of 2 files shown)


HenrikHL (changes requested):
Will break if eventLocation is null
eventLocation is used instead of vesselPosition

src/main/java/org/dcsa/core/events/service/impl/OperationsEventServiceImpl.java
```diff
@@ -17,27 +27,59 @@
...
+                .doOnNext(location -> operationsEvent.setVesselPosition(MappingUtils.instanceFrom(location, LocationTO::new, AbstractLocation.class)))
+                .thenReturn(operationsEvent);
     }
 
     @Override
     public Mono<OperationsEvent> create(OperationsEvent operationsEvent) {
         operationsEvent.setTransportCallID(operationsEvent.getTransportCall().getTransportCallID());
+        operationsEvent.setPublisher(operationsEvent.getPublisher());
         return super.save(operationsEvent);
     }
```
HenrikHL, line 84:
```suggestion
```

HenrikHL: I don't think this is needed

HenrikHL (approved):

PR #40, opened 2021-08-09
Abstracted the current impl of event subscription from TNT to event core

douradofrazer:
Resolves:
https://dcsa.atlassian.net/browse/DDT-211


src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
```diff
@@ -0,0 +1,104 @@
...
+  @GetMapping("{id}")
+  @ResponseStatus(HttpStatus.OK)
+  public Mono<T> findById(@PathVariable UUID id) {
+    return getService().findById(id);
+  }
+
+  @GetMapping
+  public Flux<T> findAll(ServerHttpResponse response, ServerHttpRequest request) {
+    ExtendedRequest<EventSubscription> extendedRequest =
+        new ExtendedRequest<>(extendedParameters, r2dbcDialect, EventSubscription.class);
```
HenrikHL, line 57:
```suggestion
        newExtendedRequest(EventSubscription.class);
```

HenrikHL: (Use the method above)

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public String getType() {
+    return "EventSubscription";
+  }
+
+  @GetMapping("{id}")
+  @ResponseStatus(HttpStatus.OK)
+  public Mono<T> findById(@PathVariable UUID id) {
+    return getService().findById(id);
+  }
```
HenrikHL, line 52: Is this needed? Isn't it the same as the base-class?

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @PostMapping(consumes = "application/json", produces = "application/json")
+  @ResponseStatus(HttpStatus.CREATED)
+  public Mono<T> create(@Valid @RequestBody T eventSubscriptionTO) {
+    if (eventSubscriptionTO.getSubscriptionID() != null) {
+      return Mono.error(new CreateException("Id not allowed when creating a new " + getType()));
+    }
+    return getService().create(eventSubscriptionTO);
+  }
```
HenrikHL, line 80: Is this needed? Isn't it the same as the BasedController (assuming the service class implements the `getIdOfEntity(...)`

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @PutMapping("{id}")
+  @ResponseStatus(HttpStatus.OK)
+  public Mono<T> update(@PathVariable UUID id, @Valid @RequestBody T eventSubscriptionTO) {
+    if (!Objects.equals(id, eventSubscriptionTO.getSubscriptionID())) {
+      return Mono.error(new UpdateException("Id in url does not match id in body"));
+    }
+    return getService().update(eventSubscriptionTO);
+  }
```
HenrikHL, line 97: Same as BaseController?

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @DeleteMapping("{id}")
+  @ResponseStatus(HttpStatus.NO_CONTENT)
+  public Mono<Void> deleteById(@PathVariable UUID id) {
+    return getService().deleteById(id);
+  }
```
HenrikHL, line 103: Same as BaseController?

src/main/java/org/dcsa/core/events/model/EventSubscription.java
```diff
@@ -0,0 +1,66 @@
...
+  private String vesselIMONumber;
+
+  @Column("transport_document_reference")
+  private String transportDocumentReference;
+
+  @Column("transport_document_type_code")
+  private String transportDocumentTypeCode;
+
+  @Column("transport_call_id")
+  private String transportCallID;
```
HenrikHL, line 39: I would have expected these fields to be added in an extension? They are very specific and not needed in all Hubs (eg: EquipmentReference is not needed in OVS-EVENT-HUB)

HenrikHL:
EventSubscriptionTransportDocumentType.java seems to be missing (is a list of either BOL or SWB)

src/main/java/org/dcsa/core/events/model/TransportDocument.java
```diff
@@ -0,0 +1,16 @@
...
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@Table("transport_document")
+public class TransportDocument extends AbstractTransportDocument {
+
+    @Column("place_of_issue")
+    private String placeOfIssue;
+
+}
```
HenrikHL, line 16: Well spotted 👍 - you removed the UUID issuer ;-)

src/main/java/org/dcsa/core/events/model/TransportDocumentType.java
```diff
@@ -0,0 +1,20 @@
+package org.dcsa.core.events.model;
+
+import lombok.Data;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.relational.core.mapping.Column;
+import org.springframework.data.relational.core.mapping.Table;
+
+@Data
+@Table("transport_document_type")
+public class TransportDocumentType {
```
HenrikHL, line 10: Is this class needed? We don't do this for any of our other enums?

src/main/java/org/dcsa/core/events/model/EventSubscription.java
```diff
@@ -0,0 +1,66 @@
...
+  private String carrierVoyageNumber;
+
+  @Column("vessel_imo_number")
+  private String vesselIMONumber;
+
+  @Column("transport_document_reference")
+  private String transportDocumentReference;
+
+  @Column("transport_document_type_code")
+  private String transportDocumentTypeCode;
```
HenrikHL, line 36: Shouldn't this be an enum?

src/main/java/org/dcsa/core/events/model/base/AbstractEventSubscription.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.data.annotation.Id;
+import org.springframework.data.relational.core.mapping.Column;
+
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotEmpty;
+import java.util.UUID;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class AbstractEventSubscription extends AuditBase {
```
HenrikHL, line 16: This class is called absctract - but it is not :-)

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+  @Query(
+          "INSERT INTO event_subscription_operations_event_type (subscription_id, operations_event_type_code)"
+                  + " VALUES (:subscriptionID, :operationsEventTypeCode)")
+  Mono<Void> insertOperationsEventTypeForSubscription(
+          UUID subscriptionID, OperationsEventTypeCode operationsEventTypeCode);
+
+  @Query(
+      "SELECT event_type FROM event_subscription_event_types"
+          + " WHERE subscription_id = :subscriptionID")
+  Flux<String> findEventTypesForSubscription(UUID subscriptionID);
```
HenrikHL, line 53: Can Spring do:
Flux<EventType> ...
mapping - or do you want it to return Flux<String>

HenrikHL: Also goes for the below methods...

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+      "SELECT equipment_event_type_code FROM event_subscription_equipment_event_type"
+          + " WHERE subscription_id = :subscriptionID")
+  Flux<String> findEquipmentEventTypesForSubscriptionID(UUID subscriptionID);
+
+  @Query(
+          "SELECT operations_event_type_code FROM event_subscription_operations_event_type"
+                  + " WHERE subscription_id = :subscriptionID")
+  Flux<String> findOperationsEventTypesForSubscriptionID(UUID subscriptionID);
+
+  @Modifying
```
HenrikHL, line 75: TransportDocumentType seems to be missing

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+import org.dcsa.core.repository.ExtendedRepository;
+import org.springframework.data.r2dbc.repository.Modifying;
+import org.springframework.data.r2dbc.repository.Query;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.util.List;
+import java.util.UUID;
+
+public interface EventSubscriptionRepository extends ExtendedRepository<EventSubscription, UUID> {
```
HenrikHL, line 14: I don't know if this comment is relevant - but wouldn't it be more efficient for *all* the below methods to take a List as parameter in order to be able to modify multiple rows in same query? I would expect for this to work - that you need to loop through your lists and delete/insert one at a time...?
Maybe not change now - but we should keep in mind for future...

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+          List<UUID> subscriptionIDs);
+
+  @Query(
+      "SELECT es.* FROM event_subscription es"
+          + " JOIN event_subscription_event_types eset"
+          + "   ON eset.subscription_id = es.subscription_id"
+          + " JOIN event_subscription_equipment_event_type eseet"
+          + "   ON eseet.subscription_id = es.subscription_id"
+          + " WHERE eset.event_type = :eventType"
+          + "    AND event_subscription_equipment_event_type.equipment_event_type_code = :equipmentEventTypeCode"
```
HenrikHL, line 140:
```suggestion
          + "   AND event_subscription_equipment_event_type.equipment_event_type_code = :equipmentEventTypeCode"
```

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+          + "   AND equipment_reference IS NULL OR equipment_reference = :equipmentReference"
+          + "   AND (carrier_booking_reference IS NULL OR carrier_booking_reference IN :carrierBookingReferences)"
+          + "   AND (transport_document_reference IS NULL OR transport_document_reference IN :transportDocumentReferences)"
+          + "   AND (transport_document_type_code IS NULL OR transport_document_type_code IN :transportDocumentTypeCodes)"
+          + "   AND (transport_call_id IS NULL OR transport_call_id = :transportCallID)"
+          + "   AND (vessel_imo_number IS NULL OR vessel_imo_number = :vesselIMONumber)"
+          + "   AND (carrier_voyage_number IS NULL OR carrier_voyage_number IN :carrierVoyageNumbers)"
+          + "   AND (carrier_service_code IS NULL OR carrier_service_code IN :carrierServiceCodes)")
+  Flux<EventSubscription> findByEquipmentEventFields(
+      EventType eventType,
```
HenrikHL, line 150: Won't this always be "EQUIPMENT" if the method is called findByEquipmentEventFields?

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+      EquipmentEventTypeCode equipmentEventTypeCode,
+      String equipmentReference,
+      List<String> carrierBookingReferences,
+      List<String> transportDocumentReferences,
+      List<String> transportDocumentTypeCodes,
+      String transportCallID,
+      String vesselIMONumber,
+      List<String> carrierVoyageNumbers,
+      List<String> carrierServiceCodes);
```
HenrikHL, line 159: Why are some of these arguments Lists while others are not?

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+      EventType eventType,
+      ShipmentEventTypeCode shipmentEventTypeCode,
+      List<String> carrierBookingReferences,
+      List<String> transportDocumentReferences,
+      List<String> transportDocumentTypeCodes,
+      List<String> transportCallIDs,
+      List<String> equipmentReferences,
+      List<String> carrierServiceCodes,
+      List<String> carrierVoyageNumbers,
+      List<String> vesselIMONumbers);
```
HenrikHL, line 187: Again - I would expect eventType to be "SHIPMENT" always when calling this method.
Why is shipmentEventTypeCode the only argument that is not a List?

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+          + "   AND (es.carrier_booking_reference IS NULL OR es.carrier_booking_reference IN :carrierBookingReferences)"
+          + "   AND (es.transport_document_reference IS NULL OR es.transport_document_reference IN :transportDocumentReferences)"
+          + "   AND (es.transport_document_type_code IS NULL OR es.transport_document_type_code IN :transportDocumentTypeCodes)"
+          + "   AND (es.transport_event_type_code IS NULL OR es.transport_event_type_code = :transportEventTypeCode)"
+          + "   AND (es.transport_call_id IS NULL OR es.transport_call_id = :transportCallID)"
+          + "   AND (es.vessel_imo_number IS NULL OR es.vessel_imo_number = :vesselIMONumber)"
+          + "   AND (es.carrier_voyage_number IS NULL OR es.carrier_voyage_number IN :carrierVoyageNumbers)"
+          + "   AND (es.carrier_service_code IS NULL OR es.carrier_service_code IN :carrierServiceCodes)"
+          + "   AND es.equipment_reference IS NULL")
+  Flux<EventSubscription> findByTransportEventFields(
```
HenrikHL, line 205: Same comment as above...

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,215 @@
...
+      List<String> carrierVoyageNumbers,
+      List<String> carrierServiceCodes,
+      TransportEventTypeCode transportEventTypeCode,
+      String vesselIMONumber,
+      String transportCallID,
+      List<String> carrierBookingReferences,
+      List<String> transportDocumentReferences,
+      List<String> transportDocumentTypeCodes,
+      EventType eventType);
+}
```
HenrikHL, line 215: findByOperationsEventFields(xxx) seems to be missing?

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
nt-as: Yes, that is doable.

src/main/java/org/dcsa/core/events/model/TransportDocumentType.java
nt-as: Agreed - we do not map these up, so we do not need this class.

src/main/java/org/dcsa/core/events/repository/ShipmentEquipmentRepository.java
```diff
@@ -0,0 +1,42 @@
+package org.dcsa.core.events.repository;
+
+import org.dcsa.core.events.model.ShipmentEquipment;
+import org.dcsa.core.repository.ExtendedRepository;
+import org.springframework.data.r2dbc.repository.Query;
+import reactor.core.publisher.Flux;
+
+import java.util.UUID;
+
+public interface ShipmentEquipmentRepository extends ExtendedRepository<ShipmentEquipment, UUID> {
```
HenrikHL, line 10: This interface is already in eBL - I am unsure if you are aware of this. It has some methods which are missing once eBL starts using EventCore.
If you think it should be updated later (when we update eBL) - this is fine with me. Just seemed like you moved other (model) classes from eBL...

src/main/java/org/dcsa/core/events/repository/TransportCallRepository.java
```diff
@@ -49,4 +49,63 @@
             + " AND facility.facility_smdg_code = :facilitySMDGCode"
             + " AND facility.un_location_code = :UNLocationCode")
     Mono<TransportCall> getTransportCall(String UNLocationCode, String facilitySMDGCode, String modeOfTransport, String vesselIMONumber);
+
+  @Query(
+      "SELECT shipping_instruction.transport_document_type FROM shipping_instruction shipping_instruction"
+          + " JOIN transport_document transport_document ON transport_document.shipping_instruction_id = shipping_instruction.id"
+          + " WHERE transport_document.transport_document_reference = :transportDocumentReference")
+  Mono<String> findTransportDocumentTypeCodeByTransportDocumentReference(
```
HenrikHL, line 57: Return `FLux<TransportDocumentType>` instead

src/main/java/org/dcsa/core/events/repository/TransportCallRepository.java
```diff
@@ -49,4 +49,63 @@
             + " AND facility.facility_smdg_code = :facilitySMDGCode"
             + " AND facility.un_location_code = :UNLocationCode")
     Mono<TransportCall> getTransportCall(String UNLocationCode, String facilitySMDGCode, String modeOfTransport, String vesselIMONumber);
+
+  @Query(
+      "SELECT shipping_instruction.transport_document_type FROM shipping_instruction shipping_instruction"
```
HenrikHL, line 54:
```suggestion
      "SELECT shipping_instruction.transport_document_type FROM shipping_instruction"
```

src/main/java/org/dcsa/core/events/repository/TransportCallRepository.java
```diff
@@ -49,4 +49,63 @@
...
+          + " JOIN transport_document transport_document ON transport_document.shipping_instruction_id = shipping_instruction.id"
+          + " WHERE transport_document.transport_document_reference = :transportDocumentReference")
+  Mono<String> findTransportDocumentTypeCodeByTransportDocumentReference(
+      String transportDocumentReference);
+
+  @Query(
+      "SELECT DISTINCT voyage.carrier_voyage_number FROM voyage voyage"
```
HenrikHL, line 61:
```suggestion
      "SELECT DISTINCT voyage.carrier_voyage_number FROM voyage"
```

src/main/java/org/dcsa/core/events/repository/TransportCallRepository.java
```diff
@@ -49,4 +49,63 @@
...
+          + " JOIN transport_call_voyage transport_call_voyage ON transport_call_voyage.voyage_id = voyage.id"
+          + " WHERE transport_call_voyage.transport_call_id = :transportCallID")
+  Flux<String> findCarrierVoyageNumbersByTransportCallID(String transportCallID);
+
+  @Query(
+      "SELECT DISTINCT service.carrier_service_code FROM service service"
```
HenrikHL, line 67: I would expect these methods to go to the ServiceRepository

src/main/java/org/dcsa/core/events/repository/TransportCallRepository.java
HenrikHL: The above is a genreal comment - i agree that sometimes you need to change the table-name - but if you don't - is it necessary to repeat the name?

src/main/java/org/dcsa/core/events/repository/TransportCallRepository.java
nt-as: As I recall, R2DBC cannot do enum mapping in this situation unfortunately (you will get a ClassCastException at runtime if you do it).

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
douradofrazer: We do here as its an ID check is required.

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
douradofrazer: You're right, removing this.

src/main/java/org/dcsa/core/events/model/EventSubscription.java
```diff
@@ -0,0 +1,63 @@
...
+  private String carrierVoyageNumber;
+
+  @Column("vessel_imo_number")
+  private String vesselIMONumber;
+
+  @Column("transport_document_reference")
+  private String transportDocumentReference;
+
+  @Column("transport_call_id")
+  private String transportCallID;
```
HenrikHL, line 36: My question still remains if this shouldn't be moved to an extended class

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
douradofrazer: Removing.

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
douradofrazer: Removing.

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
```diff
@@ -0,0 +1,239 @@
...
+
+  @Query(
+      "SELECT event_type FROM event_subscription_event_types"
+          + " WHERE subscription_id = :subscriptionID")
+  Flux<String> findEventTypesForSubscription(UUID subscriptionID);
+
+  @Query(
+      "SELECT transport_document_type_code FROM event_subscription_transport_document_type"
+          + " WHERE subscription_id = :subscriptionID")
+  Flux<String> findTransportDocumentEventTypesForSubscription(UUID subscriptionID);
```
HenrikHL, line 65: Use the Enum classes as return value instead of String - goes for all below methods as well

src/main/java/org/dcsa/core/events/model/EventSubscription.java
douradofrazer: This is a model class for the table, it will reflect all column of the event subscription table.

src/main/java/org/dcsa/core/events/model/EventSubscription.java
douradofrazer: This had been removed.

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
douradofrazer: Just added.

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
douradofrazer: Maybe it should be named as `findSubscriptionEventsByEquimentEventFields`

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
douradofrazer: It depends on the results returned by the queries executed before this, some can pin point to a singular values while other provide multiple.

src/main/java/org/dcsa/core/events/repository/ShipmentEquipmentRepository.java
douradofrazer: Was not aware of this, probably missed it, I will add those into this as well.

src/main/java/org/dcsa/core/events/repository/TransportCallRepository.java
douradofrazer: We cannot return an enum type directly, the framework does not allow it sadly.

src/main/java/org/dcsa/core/events/repository/TransportCallRepository.java
douradofrazer: Its a typo, my bad. Yup we don't need too. I just didn't see it.

src/main/java/org/dcsa/core/events/model/TransportDocumentType.java
douradofrazer: This was done inorder to have a repository for it to keep the queries related to it within that class. Plus we do have the table in the schema.

src/main/java/org/dcsa/core/events/model/base/AbstractEventSubscription.java
douradofrazer: Made it abstract, I missed it.

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
douradofrazer: Limitations of R2dbc, cannot return enum directly from the query result. There are work around but would need time to look into that.

src/main/java/org/dcsa/core/events/model/EventSubscription.java
douradofrazer: I think you mean the TO. That will be different for TNT and OVS but the model class will remain the same.

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
douradofrazer: Makes sense, also `newExtendedRequest` does not need to be generic as its implementations will always use the event subscription table. I will make that change.

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
HenrikHL: But the BaseController does this already?

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
HenrikHL: My comment was more in regard to the fact that it needs an eventType argument. But eventType is always = "EQUIPMENT" - so why not hardcode this into the query above?

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
HenrikHL: Will you accept the commit?

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
HenrikHL: Lets not to it now ;-)
LGTM

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
douradofrazer: I see it now, removing it.

src/main/java/org/dcsa/core/events/controller/AbstractEventSubscriptionController.java
douradofrazer: I added the change will commit in  a bit.

src/main/java/org/dcsa/core/events/service/impl/EventSubscriptionServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+    @Override
+    public Flux<EventSubscription> findSubscriptionsFor(Event event) {
+        switch (event.getEventType()){
+            case EQUIPMENT:
+                return findSubscriptionsFor((EquipmentEvent) event);
+            case SHIPMENT:
+                return findSubscriptionsFor((ShipmentEvent) event);
+            case TRANSPORT:
+                return findSubscriptionsFor((TransportEvent) event);
+            default:
```
HenrikHL, line 113: Operations event missing

src/main/java/org/dcsa/core/events/repository/EventSubscriptionRepository.java
douradofrazer: Added.

src/main/java/org/dcsa/core/events/service/impl/EventSubscriptionServiceImpl.java
douradofrazer: Added.

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
```diff
@@ -11,8 +11,8 @@
 
 public interface ShipmentEventRepository extends ExtendedRepository<ShipmentEvent, UUID> {
```
HenrikHL, line 12: Why do we have a ShipmentEventRepository without having any of the other EventRepositories?

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
```diff
@@ -11,8 +11,8 @@
...
 
-    @Query("SELECT * FROM shipment_event a WHERE (:eventType IS NULL or a.event_type =:eventType)")
-    Flux<ShipmentEvent> findShipmentEventsByFilters(@Param("eventType") EventType eventType);
+  @Query("SELECT * FROM shipment_event a WHERE (:eventType IS NULL or a.event_type =:eventType)")
+  Flux<ShipmentEvent> findShipmentEventsByFilters(@Param("eventType") EventType eventType);
```
HenrikHL, line 15: I am not quite sure I get this query. You are searching the Shipment_Event table for Shipment events?!? I guess you will match all for them :-D (or at least I hope you do!!)

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
```diff
@@ -11,8 +11,8 @@
...
 
   @Query(
```
HenrikHL, line 17: Why is this query here? It has nothing to do with the ShipmentEvent

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
HenrikHL: This interface seems wrong... We should discuss tomorrow

src/main/java/org/dcsa/core/events/service/impl/EventSubscriptionServiceImpl.java
```diff
@@ -0,0 +1,426 @@
...
+                return Mono.error(new UpdateException(e.getLocalizedMessage()));
+            }
+        }
+        // Ensure that the callback url at least looks valid.
+        try {
+            new URI(eventSubscription.getCallbackUrl());
+        } catch (URISyntaxException e) {
+            return Mono.error(new UpdateException("callbackUrl is invalid: " + e.getLocalizedMessage()));
+        }
+        return Mono.just(eventSubscription);
```
HenrikHL, line 103: @nt-as did we as well agree to do a HEAD call to the callbackUrl in order to validate if it exists? Or is that a TODO...

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
douradofrazer: Well , this specific method and class were written by @nt-as  :P maybe old code? I just added a few methods to it but I was.
But they all maybe redundant. I will confirm.

src/main/java/org/dcsa/core/events/service/impl/EventSubscriptionServiceImpl.java
nt-as: That is a TODO.  I think we should merge the branch without fixing that TODO first.

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
douradofrazer: FYI we do have other event repositories.

HenrikHL: My bad - thought it was a new file... I was probably too tired when I looked at this...

src/main/java/org/dcsa/core/events/repository/ShipmentEventRepository.java
douradofrazer: Removed them.

HenrikHL (approved):
Will approve this.
I do think we need to rearange @Query(..) annotations
I do think we need to remmove TransportDocumentTypeRepository (as this is done differently than all other enums)
I have not checked queries - tests will have to check if they work properly

PR #44, opened 2021-08-17
DDT-17-unofficial-APIs

Abdiiir:
Relates to DDT-17-unofficial-APIs in OVS 

- The CarrierRepo is moved to Event-Core to allow for use in VesselServiceImp
- Carrier attribute is populated

### DIFF
--- a/src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
+++ b/src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
@@ -1,0 +1,21 @@
+package org.dcsa.core.events.repository;
+
+import org.dcsa.core.repository.ExtendedRepository;
+import org.dcsa.core.events.model.Carrier;
+import org.springframework.data.r2dbc.repository.Query;
+import reactor.core.publisher.Mono;
+
+import java.util.UUID;
+
+public interface CarrierRepository extends ExtendedRepository<Carrier, UUID> {
+
+    @Query("SELECT carrier.*"
+            + "  FROM carrier"
+            + "  WHERE carrier.smdg_code =:smdgCode "
+    )
+    Mono<Carrier> getCarrierBySMdgCode(String smdgCode);
+
+    @Query("SELECT carrier FROM carrier WHERE carrier.nmfta_code =:NmftaCode")
+    Mono<Carrier> getCarrierByNmftaCode(String NmftaCode);
+}
+
--- a/src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
+++ b/src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
@@ -15,7 +15,9 @@ public interface CarrierRepository extends ExtendedRepository<Carrier, UUID> {
     )
     Mono<Carrier> getCarrierBySMdgCode(String smdgCode);
 
-    @Query("SELECT carrier FROM carrier WHERE carrier.nmfta_code =:NmftaCode")
+    @Query("SELECT carrier.*"
+            + "  FROM carrier"
+            + "  WHERE carrier.nmfta_code =:NmftaCode "
+    )
     Mono<Carrier> getCarrierByNmftaCode(String NmftaCode);
 }
-
--- a/src/main/java/org/dcsa/core/events/service/impl/TransportCallTOServiceImpl.java
+++ b/src/main/java/org/dcsa/core/events/service/impl/TransportCallTOServiceImpl.java
@@ -17,9 +17,7 @@ import org.dcsa.core.extendedrequest.ExtendedRequest;
 import org.dcsa.core.service.impl.ExtendedBaseServiceImpl;
 import org.dcsa.core.util.MappingUtils;
 import org.springframework.data.r2dbc.dialect.R2dbcDialect;
-import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
-import org.springframework.web.server.ResponseStatusException;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


src/main/java/org/dcsa/core/events/service/impl/VesselServiceImpl.java
```diff
@@ -31,8 +46,70 @@ public VesselRepository getRepository() {
...
+    public Mono<Vessel> update(Vessel vessel) {
+        return findById(getIdOfEntity(vessel))
+                .flatMap(current -> this.preUpdateHook(current, vessel))
+                .flatMap(this::save);
+    }
+
+    @Override
+    public Mono<Vessel> findById(final String VesselIMONumber) {
+        ExtendedRequest<Vessel> extendedRequest = newExtendedRequest();
+        extendedRequest.parseParameter(Map.of("vesselIMONumber", List.of(String.valueOf(VesselIMONumber))));
```
HenrikHL, line 66: @nt-as this seems somewhat overkill to me. Don't we support this in a more simple way?
(Maybe a simple vesselRepository.findById() followed by a "load-Carrier-information"...?)

src/main/java/org/dcsa/core/events/service/impl/VesselServiceImpl.java
```diff
@@ -31,8 +46,70 @@ public VesselRepository getRepository() {
...
+        switch (vessel.getVesselOperatorCarrierCodeListProvider()) {
+            case SMDG:
+                method = carrierRepository::getCarrierBySMdgCode;
+                break;
+            case NMFTA:
+                method =  carrierRepository::getCarrierByNmftaCode;
+                break;
+            default:
+                throw new CreateException("Unsupported vessel operator carrier code list provider: " + vessel.getVesselOperatorCarrierCodeListProvider());
+        }
```
HenrikHL, line 92: To me this code seems more suitable in a carrierService where you provide a codeList and a code - it then returns a carrier.
I am quite sure we need this other places than on vessel

src/main/java/org/dcsa/core/events/service/impl/VesselServiceImpl.java
nt-as: No, we currently do not support this in a more simple way without re-implementing what ExtendedRequest does for us.  Perhaps we can add something to DCSA-Core to do this after we are done with the necessary changes for the Hamburg cluster.

src/main/java/org/dcsa/core/events/service/impl/VesselServiceImpl.java
nt-as: I do not disagree that your proposal would be cleaner.  However, the proposed code works and I think we should start here and improve later.

HenrikHL: Fine with me - we just have to remember to move this code when we do the issuer on eBL (it has a issuerCode and a issuerCodeListProvider) and needs to be done in the same way...

nt-as: Hmm, if you have concrete usage case in mind already for eBL, then I am more open for it.

HenrikHL: I have this:
image

HenrikHL: Which would need the same code...

src/main/java/org/dcsa/core/events/service/impl/TransportCallTOServiceImpl.java
```diff
@@ -86,28 +84,14 @@
 
     @Override
     public Mono<TransportCallTO> create(TransportCallTO transportCallTO) {
+        transportCallTO.setVessel(transportCallTO.getVessel());
```
HenrikHL, line 87: What does this do?

HenrikHL:
Apart from my comment above this LGTM

src/main/java/org/dcsa/core/events/service/impl/TransportCallTOServiceImpl.java
nt-as: There is an assymmetri between the getters and setters.  I find it confusing as well but it is necessary right now (I will see if we can find a better solution for it).

HenrikHL (approved):

src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.UUID;
+
+public interface CarrierRepository extends ExtendedRepository<Carrier, UUID> {
+
+    @Query("SELECT carrier.*"
+            + "  FROM carrier"
+            + "  WHERE carrier.smdg_code =:smdgCode "
+    )
+    Mono<Carrier> getCarrierBySMdgCode(String smdgCode);
```
douradofrazer, line 16:
```suggestion
    Mono<Carrier> findBySmdgCode(String smdgCode);
```

src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Query("SELECT carrier.*"
+            + "  FROM carrier"
+            + "  WHERE carrier.nmfta_code =:NmftaCode "
+    )
+    Mono<Carrier> getCarrierByNmftaCode(String NmftaCode);
```
douradofrazer, line 22:
```suggestion
    Mono<Carrier> findByNmftaCode(String nmftaCode);
```

src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
```diff
@@ -0,0 +1,23 @@
+package org.dcsa.core.events.repository;
+
+import org.dcsa.core.repository.ExtendedRepository;
+import org.dcsa.core.events.model.Carrier;
+import org.springframework.data.r2dbc.repository.Query;
+import reactor.core.publisher.Mono;
+
+import java.util.UUID;
+
+public interface CarrierRepository extends ExtendedRepository<Carrier, UUID> {
```
douradofrazer, line 10: Use query methods instead of the @Query here.
You can remove `@Query(...)` if you use the suggestions mentioned below.

douradofrazer (changes requested):

src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
nt-as: https://github.com/dcsaorg/DCSA-Event-Core/pull/46

src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
nt-as: https://github.com/dcsaorg/DCSA-Event-Core/pull/46

PR #45, opened 2021-08-19
DDT-223 Fix filter on CarrierVoyageNumber and CarrierServiceCode

multimikael:
This was caused by possible NULL values. I believe the transportCallID
filter test failing was also caused by this. In addition, the voyage was
mapped different from how we mapped it in the TransportCallTO, probably 
changed information model.


multimikael:
Do I need to bump version and reference new version in https://github.com/dcsaorg/DCSA-API-Validator/pull/82 ?

nt-as (approved):

HenrikHL:
@multimikael did you enable the tests in API-Validator which we disabled in order to merge the branch we were working on into master?

PR #46, opened 2021-08-20
VesselService: Reject invalid vessel IMO in findById

nt-as:

### DIFF
--- a/src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
+++ b/src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
@@ -9,15 +9,6 @@ import java.util.UUID;
 
 public interface CarrierRepository extends ExtendedRepository<Carrier, UUID> {
 
-    @Query("SELECT carrier.*"
-            + "  FROM carrier"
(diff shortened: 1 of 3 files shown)


src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
```diff
@@ -13,11 +13,11 @@
             + "  FROM carrier"
             + "  WHERE carrier.smdg_code =:smdgCode "
     )
-    Mono<Carrier> getCarrierBySMdgCode(String smdgCode);
+    Mono<Carrier> findBySmdgCode(String smdgCode);
 
     @Query("SELECT carrier.*"
```
douradofrazer, line 18: Can remove `@Query(..)` since query method is being used. Same for above.

douradofrazer (changes requested):

src/main/java/org/dcsa/core/events/repository/CarrierRepository.java
nt-as: Done.

douradofrazer (approved):

PR #69, opened 2021-09-02
DDT-224 OVS supporting PR

multimikael:
In addition to hiding null references, it ensures the service in https://github.com/dcsaorg/DCSA-OVS/pull/41 doesn't conflict in OVS.


multimikael:
Moving to configuration variables to prevent autowired service juggling.

PR #72, opened 2021-09-03
Json ignore scheduleID

douradofrazer:


nt-as:
Let's just remove it (might require an IM change).  There is no reason to have it (we only kept it for backwards compat).

HenrikHL:
I have removed scheduleID from SwaggerHub

nt-as (approved):

PR #75, opened 2021-09-07
Added changes for new publisher object.

douradofrazer:


src/main/java/org/dcsa/core/events/model/PartyCodeListResponsibleAgency.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.UUID;
+
+@Data
+@Table("party_code_list_responsible_agency")
+public class PartyCodeListResponsibleAgency {
+
+  @Id private UUID id;
+
+  @Column("code_list_responsible_agency_code")
+  private String codeListResponsibleAgencyCode;
```
nt-as, line 17: Should this be an instance of the `CodeListResponsibleAgency` enum?

src/main/java/org/dcsa/core/events/model/base/AbstractParty.java
```diff
@@ -30,8 +30,4 @@
     @Column("public_key")
     @Size(max = 500)
     private String publicKey;
-
-    @Column("nmfta_code")
-    @Size(max = 4)
-    private String nmftaCode;
 }
```
nt-as, line 33: This is needed for backwards compat on the `TO` object (at least for eBL, unsure if we can axe it out in OVS 2.0)

src/main/java/org/dcsa/core/events/model/PartyCodeListResponsibleAgency.java
douradofrazer: The problem is that it's integers that are in string form for some reason. So until we have a dcsa version of it or something, I've kept it this way. @HenrikHL  can comment more on that.

datamodel/referencedata.d/codelistresponsibleagencycodes.csv

HenrikHL: Yes - I raised it towards @raisoman - I think we are going to make a DCSA-Version of the enum (same as ModeOfTransport)

src/main/java/org/dcsa/core/events/model/base/AbstractParty.java
HenrikHL: Yes - we need it in OVS - I deprecated the field

src/main/java/org/dcsa/core/events/repository/PartyRepository.java
```diff
@@ -9,25 +9,23 @@
 import java.util.UUID;
 
 public interface PartyRepository extends ExtendedRepository<Party, String> {
-    Mono<Party> findByAddressIDAndPartyNameAndTaxReference1AndTaxReference2AndPublicKeyAndNmftaCode(
+    Mono<Party> findByAddressIDAndPartyNameAndTaxReference1AndTaxReference2AndPublicKey(
```
HenrikHL, line 12: I am wondering if this could potentially return a Flux?

src/main/java/org/dcsa/core/events/model/base/AbstractParty.java
douradofrazer: Made the change.

src/main/java/org/dcsa/core/events/model/transferobjects/PartyTO.java
```diff
@@ -33,4 +46,35 @@ public Party toParty() {
...
+              if (null != identifyingCodes
+                      && !identifyingCodes.isEmpty()
+                      && identifyingCodes.stream()
+                      .anyMatch(
+                              idc ->
+                                      CodeListResponsibleAgency.SCAC
+                                              .getCode()
+                                              .equals(idc.getCodeListResponsibleAgencyCode()))) {
+                  throw new InvalidParameterException(
+                          "nmfta code is present along with SCAC in identifyingCodes");
```
nt-as, line 61: This would work for OVS, but I think it will hit us on eBL.

In eBL, you can update parties and that has "fun" implications for backwards compatibility (which I just realized while reviewing this).  An old client is expected to receive the identifying party codes on GET and pass them to PUT.  But since it does not what it contains or how to deal with it, the client will have to forward it as-is.  At the same time, it _is_ allowed to change the `nmftaCode` _without_ understanding the identifying party codes field (otherwise, this breaks backwards compatibility).

In summary, I think we need to handle this as:

- if nmftaCode is set, then it overrules the SCAC identifying party code on conflict.  This implies that a new client must either clear the nmftaCode field _OR_ update nmftaCode to match the SCAC identifying party code as well (or only update the nmftaCode field).
- if nmftaCode is not set, then the SCAC code from the identifying party codes applies unconditionally (if present).

src/main/java/org/dcsa/core/events/repository/PartyRepository.java
nt-as: It would be now, yes.  You can (in theory) have different identifying party codes with the same set of values and that would be two distinct entities.

Otherwise, we would "leak" party codes between clients that share the same base party object (notably `zzz` codes are going to cause issues here as the clients would have a different interpretation of these).

src/main/java/org/dcsa/core/events/service/impl/PartyServiceImpl.java
```diff
@@ -41,10 +47,30 @@ public PartyRepository getRepository() {
...
+            .flatMap(
+                pTo ->
+                    Util.createOrFindByContent(
+                        pTo, partyRepository::findByContent, pTO -> this.create(pTO.toParty())))
+            .flatMap(
+                party ->
+                    Flux.fromIterable(
+                            partyTO.getIdentifyingCodes().stream()
+                                .map(idc -> mapIdCodeToPartyCLRA.apply(party.getId(), idc))
+                                .collect(Collectors.toList()))
```
nt-as, line 59: Nit: Flux has an "fromStream" method which means we can omit the `.collect(Collectors.toList())` call.

nt-as: Alternatively, use `fromIterable` directly on `partyTO.getIdentifyingCodes()` and then use the standard `.map` operator on the flux.

src/main/java/org/dcsa/core/events/repository/PartyRepository.java
nt-as: I think we will need a different way to de-duplicate these now as a side-effect :-/

src/main/java/org/dcsa/core/events/model/transferobjects/PartyTO.java
```diff
@@ -33,4 +45,42 @@ public Party toParty() {
...
+                              idc ->
+                                      CodeListResponsibleAgency.SCAC
+                                              .getCode()
+                                              .equals(idc.getCodeListResponsibleAgencyCode()))) {
+
+                  for (IdentifyingCode idc : this.identifyingCodes) {
+                      if(CodeListResponsibleAgency.SCAC
+                              .getCode()
+                              .equals(idc.getCodeListResponsibleAgencyCode())){
+                            idc.setPartyCode(this.getNmftaCode());
```
nt-as, line 64: Nit: `break;` would be fine here (can come later)

nt-as: No, I misread. Code is correct.

nt-as (approved):

PR #77, opened 2021-09-08
GenericEventService: Make some fields visible for subclasses

nt-as:

### DIFF
--- a/src/main/java/org/dcsa/core/events/service/impl/GenericEventServiceImpl.java
+++ b/src/main/java/org/dcsa/core/events/service/impl/GenericEventServiceImpl.java
@@ -6,16 +6,21 @@ import org.dcsa.core.events.model.*;
 import org.dcsa.core.events.model.enums.EventType;
 import org.dcsa.core.events.repository.EventRepository;
 import org.dcsa.core.events.service.*;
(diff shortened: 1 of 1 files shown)


douradofrazer (approved):

src/main/java/org/dcsa/core/events/service/impl/GenericEventServiceImpl.java
```diff
@@ -30,41 +35,79 @@ public EventRepository getRepository() {
 
     @Override
     public Flux<Event> findAllExtended(ExtendedRequest<Event> extendedRequest) {
-        return super.findAllExtended(extendedRequest);
+        Set<EventType> eventTypes = getSupportedEvents();
+        return super.findAllExtended(extendedRequest)
+                // TODO: Push this filter into the extendedRequest, so the database does not spend time on them.
```
HenrikHL, line 40: But is being done already... In the subclasses the system adds to eventType query paraemter if it is not added by the user - so this should not be needed

nt-as: No, they do not.  They filter it out in a similar matter (though via a `flatMap` after the `findAllExtended`). :)

PR #80, opened 2021-09-09
Bump core

nt-as:


nt-as:
`getTimestampTypeName()` was added because both DCSA-OVS-Notifications and DCSA-UI-Support needs that code.

HenrikHL (approved):

PR #94, opened 2021-09-27
Event Type filter for GET event-subscriptions.

douradofrazer:


src/main/java/org/dcsa/core/events/model/base/AbstractEventSubscription.java
```diff
@@ -26,4 +29,8 @@
   @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
   // Jackson encodes this in base64 by default
   protected byte[] secret;
+
+  public List<EventType> getEventType() {
+    return Collections.emptyList();
+  }
```
nt-as, line 35: I think we should let this be abstract.  The default implementation is "never" useful.

nt-as (approved):

PR #95, opened 2021-09-28
DDT-392: Ignore facility code when locating TransportCalls

nt-as:
The facility code is not a part of the (logical) transport call as
clarified today with Nicolas and Charles (see DDT-392) when
negotiating timestamps. For a given "Port of Call", a vessel can move
between different facilities and it should still be considered the
same `TransportCall`.


nt-as:
mmm, let me just come back to you on this one.

douradofrazer (approved):

PR #96, opened 2021-09-30
Bumped core to latest in event core, to use profile configs.

douradofrazer:


nt-as:
Can we hold this? I got some pending changes for the event core that I need to include for OVS.

nt-as (approved):

PR #97, opened 2021-09-30
DDT-392: Move facility to location

nt-as:
... Also include some of the new JIT 1.1 timestamps

### DIFF
--- a/src/main/java/org/dcsa/core/events/model/OperationsEvent.java
+++ b/src/main/java/org/dcsa/core/events/model/OperationsEvent.java
@@ -21,6 +21,8 @@ import org.springframework.data.relational.core.mapping.Table;
 @JsonTypeName("OPERATIONS")
 public class OperationsEvent extends Event implements TransportCallBasedEvent {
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/dcsa/core/events/model/OperationsEvent.java
```diff
@@ -76,30 +78,69 @@
...
+        }
+        if (portCallServiceTypeCode != null) {
+            // Special-cases around FAST, Gangway and SAFE
+            switch (portCallServiceTypeCode) {
+                case FAST:
+                    if (classifierCode != EventClassifierCode.ACT || facilityTypeCode != FacilityTypeCode.BRTH
+                            || operationsEventTypeCode != OperationsEventTypeCode.ARRI) {
+                        return UNKNOWN_TIMESTAMP;
+                    }
+                    return "AT All fast";
```
HenrikHL, line 109:
```suggestion
                    return "All fast";
```

nt-as: No, it is literally called `AT All fast`.

HenrikHL (approved):

PR #103, opened 2021-10-06
Added enums needed for booking summaries.

douradofrazer:


src/main/java/org/dcsa/core/events/model/enums/PaymentTerm.java
```diff
@@ -0,0 +1,16 @@
+package org.dcsa.core.events.model.enums;
+
+public enum PaymentTerm {
+  PRE("Prepaid"),
+  COL("Collect");
+
+  private final String value;
```
nt-as, line 7: Use `@Getter` from Lombok?

douradofrazer:
Note: I will not be merging this branch till booking development is done, so I will use the branch snapshot to build DCSA-BKG.

src/main/java/org/dcsa/core/events/model/enums/DocumentStatus.java
```diff
@@ -0,0 +1,24 @@
+package org.dcsa.core.events.model.enums;
+
+public enum DocumentStatus {
```
HenrikHL, line 3: I am thinking this enum-class can be used as a base-class for ShipmentEventTypeCode enum. Please confirm @raisoman ...
Will DocumentStatus **always** be a subset of ShipmentEventTypeCode?

PR #123, opened 2021-11-04
DDT-520-Allow for extraction of timestamp definition using EventID

Abdiiir:
A Query is created to support extraction timestamp type name for issue DDT-520


nt-as:
Needs version bump

nt-as (approved):

PR #124, opened 2021-11-08
Shipment entity + repository

mp-assco:


src/main/java/org/dcsa/core/events/model/Shipment.java
```diff
@@ -0,0 +1,40 @@
...
+
+import javax.validation.constraints.Size;
+import java.time.OffsetDateTime;
+import java.util.UUID;
+
+@Data
+@Table("shipment")
+public class Shipment {
+
+  @JsonIgnore
```
douradofrazer, line 17: `@JsonIgnore` is not required here as we will not be using it as a TO.

src/main/java/org/dcsa/core/events/model/Shipment.java
```diff
@@ -0,0 +1,40 @@
...
+  @Column("booking_id")
+  private UUID bookingID;
+
+  @JsonIgnore
```
douradofrazer, line 21: Ditto

src/main/java/org/dcsa/core/events/model/Shipment.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Size(max = 35)
+  @Column("carrier_booking_reference")
+  private String carrierBookingReferenceID;
+
+  @Column("terms_and_conditions")
+  private String termsAndConditions;
+
+  @Column("confirmation_datetime")
+  @JsonFormat(shape = JsonFormat.Shape.STRING)
```
douradofrazer, line 33: `@JsonFormat` not required for the same reason.

src/main/java/org/dcsa/core/events/model/Shipment.java
```diff
@@ -0,0 +1,40 @@
...
+  private OffsetDateTime confirmationDateTime;
+
+  @JsonIgnore
```
douradofrazer, line 36: Ditto

douradofrazer (changes requested):

douradofrazer (approved):

PR #126, opened 2021-11-08
DDT-362 aligned booking model with latest IM changes

pimg:


src/main/java/org/dcsa/core/events/model/Booking.java
```diff
@@ -38,20 +41,80 @@
   private CargoMovementType cargo_movement_type_at_destination;
```
douradofrazer, line 41: We should rename this as `cargoMovementTypeAtOrigin` instead of `cargo_movement_type_at_destination` to keep things consistent.

douradofrazer: You might see such inconsistencies , if so feel free to correct them when you come across them.

src/main/java/org/dcsa/core/events/model/Booking.java
```diff
@@ -38,20 +41,80 @@
...
 
   @Column("booking_request_datetime")
-  private OffsetDateTime bookingDateTime;
+  private OffsetDateTime bookingRequestDateTime;
 
-  @Column("service_contract")
+  @Column("service_contract_reference")
   @Size(max = 30)
   private String serviceContract;
```
douradofrazer, line 48: `serviceContract` -> `serviceContractReference`

src/main/java/org/dcsa/core/events/model/Booking.java
```diff
@@ -38,20 +41,80 @@
...
+  private String transportDocumentReference;
+
+  @Column("booking_document_reference")
+  @Size(max = 20)
+  private String bookingDocumentReference;
+
+  @Column("booking_channel_reference")
   @Size(max = 20)
-  private String commodityType;
+  private String bookingChannelReference;
```
douradofrazer, line 105: This isn't in the IM SQL, is it in the doc?

src/main/java/org/dcsa/core/events/model/Booking.java
```diff
@@ -38,20 +41,80 @@
...
+
+  @Column("communication_channel_code")
+  private CommunicationChannel communicationChannelCode;
+
+  @Column("is_equipment_substitution_allowed")
+  private Boolean isEquipment_SubstitutionAllowed;
```
douradofrazer, line 111: `isEquipment_SubstitutionAllowed` -> `isEquipmentSubstitutionAllowed`

src/main/java/org/dcsa/core/events/model/Booking.java
```diff
@@ -38,20 +41,80 @@
...
+
+  @Column("transport_document_type_code")
+  private TransportDocumentTypeCode transportDocumentTypeCode;
+
+  @Column("transport_document_reference")
+  private String transportDocumentReference;
+
+  @Column("booking_document_reference")
+  @Size(max = 20)
+  private String bookingDocumentReference;
```
douradofrazer, line 101: This isn't in the IM SQL, is it in the doc?

src/main/java/org/dcsa/core/events/model/Booking.java
```diff
@@ -21,14 +21,17 @@
 public class Booking extends AuditBase {
 
   @Id
-  @Column("carrier_booking_reference")
-  @Size(max = 35)
+  @Column("carrier_booking_request_reference")
+  @Size(max = 100)
   private String id;
```
douradofrazer, line 26: The `id` and `carrier_booking_request_reference` are two different columns. `id` is the primary key and the other is a reference that will be exposed.

douradofrazer (changes requested):

src/main/java/org/dcsa/core/events/model/Booking.java
pimg: it is in the IM SQL: https://github.com/dcsaorg/DCSA-Information-Model/blob/IM_3.5_alignment/datamodel/initdb.d/03_dcsa_im_v3_0.sql#L224

douradofrazer (approved):
LGTM!

PR #134, opened 2021-11-15
Use Vessel ID instead of IMO

mp-assco:


nt-as (changes requested):
This commit is incomplete.  Please compare it with commit 56035357 (which was a generic bump to IM 3.4 and therefore covers this update)

mp-assco:
Apparently this has already been implemented in master as Niels mentioned in his comment. Suggest we either cherry-pick https://github.com/dcsaorg/DCSA-Event-Core/commit/5603535730379ac652382a4983bb1e68833cc565 or merge master into dev branch.

PR #136, opened 2021-11-15
Added entities and repositories to support ticket DDT-365

douradofrazer:
**Only review the Repositories and Entities.**

Note: Also rebased with the current master since it was out of date.


src/main/java/org/dcsa/core/events/model/NegotiationCycle.java
```diff
@@ -0,0 +1,17 @@
+package org.dcsa.core.events.model;
+
+import lombok.Data;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.relational.core.mapping.Column;
+import org.springframework.data.relational.core.mapping.Table;
+
+@Table("negotiation_cycle")
```
pimg, line 8: This table is not present in IM 3.5 currently

src/main/java/org/dcsa/core/events/model/OperationsEvent.java
```diff
@@ -14,28 +14,32 @@
...
+
 @Table("operations_event")
 @NoArgsConstructor
 @Data
 @EqualsAndHashCode(callSuper = true)
 @JsonTypeName("OPERATIONS")
 public class OperationsEvent extends Event implements TransportCallBasedEvent {
 
-    public static final String UNKNOWN_TIMESTAMP = "<Unknown timestamp>";
-
```
pimg, line 25: missing publisher

src/main/java/org/dcsa/core/events/model/TimestampDefinition.java
```diff
@@ -0,0 +1,71 @@
...
+    @Column("provided_in_standard")
+    private String providedInStandard;
+
+    @Column("accept_timestamp_definition")
+    private String acceptTimestampDefinition;
+
+    @Column("reject_timestamp_definition")
+    private String rejectTimestampDefinition;
+
+    @Column("canonical_timestamp_definition")
```
pimg, line 69: this column is not in the 3.5 IM supporting tables

src/main/java/org/dcsa/core/events/ShipmentLocation.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.data.relational.core.mapping.Column;
+import org.springframework.data.relational.core.mapping.Table;
+
+import java.time.OffsetDateTime;
+import java.util.UUID;
+
+@Data
+@Table("shipment_location")
+public class ShipmentLocation {
+  @Id private UUID id;
```
mp-assco, line 15: Should this be tagged as `JsonIgnore`?

src/main/java/org/dcsa/core/events/ShipmentLocation.java
```diff
@@ -0,0 +1,34 @@
...
+  private UUID bookingID;
+
+  @Column("location_id")
+  private String locationID;
+
+  @Column("shipment_location_type_code")
+  private LocationType shipmentLocationTypeCode;
+
+  @Column("displayed_name")
+  private String displayedName;
```
mp-assco, line 30: Required field. Don't know if we're bothering with this at this moment, though?

src/main/java/org/dcsa/core/events/model/NegotiationCycle.java
nt-as: (For reference to other reviewers, it is a implementation detail in the reference implementation and therefore present in the 05 SQL file rather than the official SQL/IM)

src/main/java/org/dcsa/core/events/model/TimestampDefinition.java
nt-as: It was added in commit https://github.com/dcsaorg/DCSA-Information-Model/commit/e614fc97ecdf409c782b1c04bc9fe8702f876d88

pimg (commented):
left one comment

src/main/java/org/dcsa/core/events/ShipmentLocation.java
douradofrazer: We have a DTO object now in Booking.  This is purely a model/entity

src/main/java/org/dcsa/core/events/ShipmentLocation.java
douradofrazer: Validations are now on the DTO object

pimg (approved):

PR #139, opened 2021-11-16
Move ShipmentLocation

mp-assco:


src/main/java/org/dcsa/core/events/model/AbstractShipmentLocation.java
```diff
@@ -1,34 +1,47 @@
...
+import javax.validation.constraints.Size;
 import java.time.OffsetDateTime;
 import java.util.UUID;
 
 @Data
+@NoArgsConstructor
 @Table("shipment_location")
-public class ShipmentLocation {
-  @Id private UUID id;
+public abstract class AbstractShipmentLocation {
```
douradofrazer, line 20: We want to keep the model completely separate from the TO, why do we need  @JsonIgnore and Validations here?

douradofrazer (changes requested):

src/main/java/org/dcsa/core/events/model/AbstractShipmentLocation.java
douradofrazer: TO object is already present in DCSA-BKG

mp-assco: I thought the plan was to extend the TO in the Booking service with the existing entity.

douradofrazer (approved):

PR #145, opened 2021-11-22
DDT-382: Carrier Clauses

mp-assco:
Repository for both CarrierClause and ShipmentCarrierClause


src/main/java/org/dcsa/core/events/model/CarrierClause.java
```diff
@@ -12,7 +12,8 @@
 public class CarrierClause {
 
   @Id
-  private UUID id;
+  @Column("id")
+  private UUID carrierClauseID;
```
pimg, line 16: The field in the IM is named id, why was it renamed to carrierClauseID in the model Object?

pimg (changes requested):
carrierClauseID seems to mismatch the IM

src/main/java/org/dcsa/core/events/model/CarrierClause.java
mp-assco: Had an ambiguous method issue in testing for `any()`, but could just have cast to `UUID`. 

I have changed back.

pimg (approved):

PR #154, opened 2021-11-29
DDT-608: Reference rename

mp-assco:


HenrikHL (dismissed):
The name doesn't have to be escaped any more I guess - but no harm done...

src/main/java/org/dcsa/core/events/repository/ServiceRepository.java
```diff
@@ -39,7 +39,7 @@
           + " ON st.transport_id = t.id "
           + "JOIN cargo_item ci "
           + " ON ci.shipment_id = st.shipment_id "
-          + "LEFT JOIN \"references\" r"
+          + "LEFT JOIN \"reference\" r"
           + " ON r.shipment_id = st.shipment_id"
           + "WHERE (ci.shipping_instruction_id = :shippingInstructionID OR references.shipping_instruction_id = :shippingInstructionID)")
```
HenrikHL, line 44:
```suggestion
          + "WHERE (ci.shipping_instruction_id = :shippingInstructionID OR reference.shipping_instruction_id = :shippingInstructionID)")
```

src/main/java/org/dcsa/core/events/repository/VoyageRepository.java
```diff
@@ -32,7 +32,7 @@
           " ON st.transport_id = t.id " +
           "JOIN cargo_item ci " +
           " ON ci.shipment_id = st.shipment_id " +
-          "LEFT JOIN \"references\" r" +
+          "LEFT JOIN \"reference\" r" +
           " ON r.shipment_id = st.shipment_id" +
           "WHERE (ci.shipping_instruction_id = :shippingInstructionID OR references.shipping_instruction_id = :shippingInstructionID)")
```
HenrikHL, line 37:
```suggestion
          "WHERE (ci.shipping_instruction_id = :shippingInstructionID OR reference.shipping_instruction_id = :shippingInstructionID)")
```

HenrikHL:
I found 2 places in this PR that needs renaming. Please make a global search - there might be other places as well...

nt-as:
> The name doesn't have to be escaped any more I guess - but no harm done...

I would say we should remove the quotes because it _can_ cause issues.  By default, the database case normalizes unquoted input and accordingly the "public wisdom" is to either _always_ quote or _never_ quote.  Except for `"refereces"` we have never quoted, so I think we should standardize on "not quoting" now that we have the option (this implies that we should not quote the name in the SQL either in case we did that).

mp-assco:
> > The name doesn't have to be escaped any more I guess - but no harm done...
> 
> I would say we should remove the quotes because it _can_ cause issues. By default, the database case normalizes unquoted input and accordingly the "public wisdom" is to either _always_ quote or _never_ quote. Except for `"refereces"` we have never quoted, so I think we should standardize on "not quoting" now that we have the option (this implies that we should not quote the name in the SQL either in case we did that).

@nt-as I have removed the quotations.

HenrikHL (approved):

PR #167, opened 2021-12-16
Created and updated date time booking

mp-assco:


pimg (approved):

src/main/java/org/dcsa/core/events/model/Booking.java
```diff
@@ -131,4 +132,13 @@
   public boolean isNew() {
     return this.getId() == null;
   }
-}
+
+  // Both eventCreatedDateTime and eventDateTime are considered metadata
+  // and are not part of the official IM model. They are added in the sql only.
+
+  @Column("created_date_time")
```
HenrikHL, line 139: @pimg we already have this in line 50 (bookingRequestDateTime) - we don't need it twice ;-)

PR #180, opened 2022-01-19
Bumped core and event core ver.

douradofrazer:


douradofrazer:
Not required, as the update is made in dev-bkg.

PR #183, opened 2022-01-26
Provide standardized validation for location types

nt-as:
This validation is for when the API requests mandates a particular
type of location such as `addressLocation` or `facilityLocation`.  In
this case, it can be used as the following:

```java
   @ValidLocationSubtype(anyOf = {ValidLocationSubtype.LocationSubtype.ADDRESS, ValidLocationSubtype.LocationSubtype.FACILITY})
   private final LocationTO addressOrFacilityLocationType;

```

Note that the annotation provides a minimum validation and permits
other fields to be present as well.  Accordingly, a geo coordinate
with an address would count as both a `GEO_COORDINATE` and an
`ADDRESS` location at the same time and would pass the restriction
from the example.

In the rare case that `and` restrictions are needed, then the
annotation can be repeated.  It will still not prevent "additional"
fields but can be useful to ensure that you have a facility with a UN
Location Code (to future proof in case we get facility code lists that
do not depend on UN Location codes).  As an example:

```java
     // Either UN Location Code OR facility (or both)
     @ValidLocationSubtype(anyOf = {ValidLocationSubtype.LocationSubtype.UN_LOCATION, ValidLocationSubtype.LocationSubtype.FACILITY})
     private final LocationTO poteLocation;

     // UN Location code AND facility
     @ValidLocationSubtype(anyOf = ValidLocationSubtype.LocationSubtype.UN_LOCATION)
     @ValidLocationSubtype(anyOf = ValidLocationSubtype.LocationSubtype.FACILITY)
     private final LocationTO portWithTerminalLocation;
```

### DIFF
--- a/src/main/java/org/dcsa/core/events/validator/ValidLocationSubtype.java
+++ b/src/main/java/org/dcsa/core/events/validator/ValidLocationSubtype.java
@@ -1,0 +1,106 @@
+package org.dcsa.core.events.validator;
+
+import org.dcsa.core.events.model.transferobjects.LocationTO;
+
+import javax.validation.Constraint;
+import javax.validation.Payload;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.*;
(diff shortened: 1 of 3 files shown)


HenrikHL:
LGTM

mp-assco (approved):
looks good to me

PR #193, opened 2022-02-01
DDT-761 insert shipping equipment

pimg:
resolves DDT-761


src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
```diff
@@ -0,0 +1,65 @@
...
+import org.dcsa.core.model.GetId;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.relational.core.mapping.Column;
+
+import javax.validation.constraints.Size;
+import java.util.UUID;
+
+@NoArgsConstructor
+@Data
+public class ActiveReeferSettings extends AuditBase implements GetId<UUID> {
```
nt-as, line 18:
```suggestion
public class ActiveReeferSettings extends AuditBase {
```

`GetId<UUID>` is generally redundant.  Is there a particular use for it here (otherwise I would prefer we removed it).

(There are tons of cases of `GetId<...>` sprinkled in the code base for historical reasons. Lets avoid spreading it further)

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Column("ventilation_max")
+    private Float ventilationMax;
+
+    @Override
+    @JsonIgnore
+    public UUID getId() {
+        return getShipmentEquipmentID();
+    }
+
```
nt-as, line 53: This can probably go as well.  Calls can often just be replaced by `getShipmentEquipmentID()`

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
```diff
@@ -0,0 +1,65 @@
...
+    public ActiveReeferSettingsTO toActiveReeferSettingsTO() {
```
nt-as, line 54: Should this be done by a Mapper instead?

src/main/java/org/dcsa/core/events/model/CargoItem.java
```diff
@@ -4,31 +4,71 @@
...
+  private String shippingInstructionID;
 
-    @Column("shipment_equipment_id")
-    @NotNull
-    protected UUID shipmentEquipmentID;
+  @Column("shipment_equipment_id")
+  @NotNull
+  protected UUID shipmentEquipmentID;
+
+  public CargoItemTO toCargoItemTO() {
```
nt-as, line 37: Ditto (Should this be implemented via a mapper instead)?

Note: This seems to be a reoccurring thing, so I will not comment on further instances in this review.

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
```diff
@@ -0,0 +1,150 @@
...
+                            Mono.zip(
+                                Mono.just(shipmentEquipment.getId()),
+                                Mono.just(shipmentEquipmentTO))));
+  }
+
+  private Mono<EquipmentTO> saveEquipment(EquipmentTO equipmentTO) {
+    if (Objects.isNull(equipmentTO)) {
+      return Mono.empty();
+    }
+    return equipmentRepository.save(equipmentTO.toEquipment()).map(Equipment::toEquipmentTO);
```
nt-as, line 86:
```suggestion
    return Mono.justOrEmpty(equimentTO)
             .map(equipmentTO::toEquipment)
             .flatMap(equipmentRepository::save)
             .map(Equipment::toEquipmentTO);
```

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  private Mono<ActiveReeferSettingsTO> saveActiveReeferSettings(
+      UUID shipmentEquipmentID, ActiveReeferSettingsTO activeReeferSettingsTO) {
+    if (Objects.isNull(activeReeferSettingsTO)) {
+      return Mono.empty();
+    }
+    return activeReeferSettingsRepository
+        .save(activeReeferSettingsTO.toActiveReeferSettings(shipmentEquipmentID))
+        .map(ActiveReeferSettings::toActiveReeferSettingsTO);
```
nt-as, line 96: Ditto, the `if` can be replaced by a `Mono.justOrEmpty`  + map/flatMap.

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
```diff
@@ -0,0 +1,150 @@
...
+        .map(Seal::toSealTO)
+        .collectList();
+  }
+
+  private Mono<List<CargoItemTO>> saveCargoItems(
+      UUID shipmentEquipmentID, String shippingInstructionID, List<CargoItemTO> cargoItemTOs) {
+    if (Objects.isNull(cargoItemTOs) || cargoItemTOs.isEmpty()) {
+      return Mono.just(Collections.emptyList());
+    }
+    return Flux.fromStream(cargoItemTOs.stream())
```
nt-as, line 114:
```suggestion
    return Flux.fromIterable(cargoItemTOs)
```

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
pimg: fixed

src/main/java/org/dcsa/core/events/model/CargoItem.java
pimg: fixed all to..() methods have been replaced by Mappers

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
pimg: fixed

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
pimg: fixed

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
pimg: fixed

src/main/java/org/dcsa/core/events/model/Equipment.java
```diff
@@ -27,4 +30,13 @@
...
   private String weightUnit;
+
+  public EquipmentTO toEquipmentTO() {
+    EquipmentTO equipmentTO = new EquipmentTO();
+    equipmentTO.setEquipmentReference(this.getEquipmentReference());
+    equipmentTO.setIsoEquipmentCode(this.getIsoEquipmentCode());
+    equipmentTO.setTareWeight(this.getTareWeight());
+    equipmentTO.setWeightUnit(WeightUnit.valueOf(this.getWeightUnit()));
+    return equipmentTO;
+  }
```
nt-as, line 41: Mapper?

nt-as (commented):
There are several places where we use `flatMap` on a `Flux` in this PR.  While it does not look problematic per-se in this case, we have lot of time in the past debugging why sorting does not work because someone used an unfortunately placed `flatMap` instead of `concatMap`.

An open question here is then, should we prefer `concatMap` to prime future changes to use `concatMap` in other to reduce the risk of this kind of problem or are we fine with `flatMap` even if it occasionally causes a sorting bug due to `copy-waste` (whether metaphorical or literal copy-paste)?

(If we prefer `concatMap` then I will try to go through the code one more time to find `Flux.flatMap` calls, otherwise, I will leave it at this)

src/main/java/org/dcsa/core/events/model/Equipment.java
pimg: Ah missed one to..() to remove, this is fixed now. 
As far as the flatMap vs concatMap. I decided to keep the flatMap, since this is een create (insert) method the sorting becomes irrelevant. I think it is not necessary to only use concatMap or flatMap.

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.validation.constraints.Size;
+import java.util.UUID;
+
+@NoArgsConstructor
+@Data
+public class ActiveReeferSettings extends AuditBase {
+
+    @Id
+    /* We do not show this in JSON as it is an internal detail */
+    @JsonIgnore
```
nt-as, line 20: Given we have now made an `ActiveReeferSettingsTO` then it seems weird to have `@JsonIgnore` on the DAO variant (as it will not be exposed as JSON in the first place).

src/main/java/org/dcsa/core/events/model/transferobjects/ActiveReeferSettingsTO.java
```diff
@@ -0,0 +1,25 @@
...
+import javax.validation.constraints.Size;
+
+@Data
+public class ActiveReeferSettingsTO {
+
+	private Float temperatureMin;
+
+	private Float temperatureMax;
+
+	@Size(max = 3)
```
nt-as, line 15: We probably do not need a `@Size` here given it is for an enum (which dictates exact values anyway).

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
```diff
@@ -0,0 +1,152 @@
...
+        .flatMap(
+            cargoItemTO ->
+                referenceService
+                    .createReferencesByShippingInstructionIDAndTOs(
+                        shippingInstructionID, cargoItemTO.getReferences())
+                    .thenReturn(cargoItemTO))
+        .collectList();
+  }
+
+  private Mono<CargoItemTO> saveCargoLineItems(UUID cargoItemId, CargoItemTO cargoItemTO) {
```
nt-as, line 134: Nit: `cargoItemID`

nt-as (approved):
LGTM - had a few final nits, but I do not think I need to re-review it after they have been resolved.

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
pimg: fixed

src/main/java/org/dcsa/core/events/model/transferobjects/ActiveReeferSettingsTO.java
pimg: fixed

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
pimg: fixed

nt-as (approved):

PR #203, opened 2022-02-10
ShipmentEquipment and Seal according to IM

mp-assco:
SealMapper did not set ShipmentEquipmentID


HenrikHL:
Looks good to me...

src/main/java/org/dcsa/core/events/model/Seal.java
```diff
@@ -24,18 +25,22 @@
...
 
     @Column("seal_number")
     @Size(max = 15)
+    @NotNull
     private String sealNumber;
 
-    @Column("seal_source")
+    @Column("seal_source_code")
     @Size(max = 5)
+    @NotNull
```
HenrikHL, line 38: This is not correct!

HenrikHL: This field needs to be optional

PR #205, opened 2022-02-10
Align ActiveReeferSettings, CargoItem, and Ref with IM

mp-assco:


src/main/java/org/dcsa/core/events/model/Reference.java
```diff
@@ -26,6 +27,7 @@
   private UUID referenceID;
 
   @Column("reference_type_code")
+  @NotNull
   private ReferenceTypeCode referenceType;
 
   @Size(max = 100)
```
HenrikHL, line 33: We need a @NotNull here

mp-assco: Done

HenrikHL (dismissed):
LGTM

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
```diff
@@ -1,24 +1,36 @@
...
+import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
 import java.util.UUID;
 
 @NoArgsConstructor
 @Data
-public class ActiveReeferSettings extends AuditBase {
+@Table("active_reefer_settings")
+@EqualsAndHashCode(callSuper = true)
+public class ActiveReeferSettings extends AuditBase implements Persistable<UUID> {
```
douradofrazer, line 23: Could you remove AuditBase?  we do not use it.

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
```diff
@@ -1,24 +1,36 @@
...
 
     @Id
     /* We do not show this in JSON as it is an internal detail */
     @Column("shipment_equipment_id")
     private UUID shipmentEquipmentID;
 
+    @Transient
+    @JsonIgnore
```
douradofrazer, line 31: No need for @JsonIgnore,  we have separated entities and DTOs

douradofrazer (changes requested):

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
mp-assco: Done

src/main/java/org/dcsa/core/events/model/ActiveReeferSettings.java
mp-assco: Done

pimg (approved):

PR #207, opened 2022-02-11
Preperation for slimming DCSA-Core

nt-as:
Contains:

1. A left over GetId usage.
2. Update of most repositories to cope with ExtendedRepository becoming a fragment.
3. Update of most services to cope with BaseService (and all of its default methods) being removed.

Note that 2+3 are incomplete and only covers some of the "trivial" cases. There will be a follow up for these.


mp-assco (approved):
lgtm

mp-assco (dismissed):
slgtm

HenrikHL (dismissed):

pimg (approved):

PR #212, opened 2022-02-15
Reduce the use of DCSA-Core in cases where it is not necessary

nt-as:
These commits will enable the removal/scope reduction of:

- `BaseService` (removal)
- `ExtendedBaseService` reduced to `findAllExtended`
- `BaseController` (removal)
- `ExtendedBaseController` reduced a `@GetMapping` for getting all entities

To keep things easier to review, I have excluded the changes to repositories for now as they can come in a separate PR.

### DIFF
--- a/src/main/java/org/dcsa/core/events/service/FacilityService.java
+++ b/src/main/java/org/dcsa/core/events/service/FacilityService.java
@@ -8,10 +8,12 @@ import reactor.core.publisher.Mono;
 import java.util.UUID;
 
 public interface FacilityService extends ExtendedBaseService<Facility, UUID> {
+
   Mono<Facility> findByUNLocationCodeAndFacilityCode(
       String unLocationCode,
       FacilityCodeListProvider facilityCodeListProvider,
       String facilityCode);
 
   Mono<Facility> findByIdOrEmpty(UUID id);
+  Mono<Facility> findById(UUID uuid);
 }
--- a/src/main/java/org/dcsa/core/events/service/TimestampDefinitionService.java
+++ b/src/main/java/org/dcsa/core/events/service/TimestampDefinitionService.java
@@ -3,9 +3,12 @@ package org.dcsa.core.events.service;
 import org.dcsa.core.events.model.OperationsEvent;
 import org.dcsa.core.events.model.TimestampDefinition;
 import org.dcsa.core.service.ExtendedBaseService;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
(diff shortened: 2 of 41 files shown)


src/main/java/org/dcsa/core/events/service/FacilityService.java
```diff
@@ -8,10 +8,12 @@
 import java.util.UUID;
 
 public interface FacilityService extends ExtendedBaseService<Facility, UUID> {
```
pimg, line 10: shouldn't the extends ExtendedBaseService be removed?

src/main/java/org/dcsa/core/events/service/TimestampDefinitionService.java
```diff
@@ -3,9 +3,12 @@
 import org.dcsa.core.events.model.OperationsEvent;
 import org.dcsa.core.events.model.TimestampDefinition;
 import org.dcsa.core.service.ExtendedBaseService;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 public interface TimestampDefinitionService extends ExtendedBaseService<TimestampDefinition, String> {
```
pimg, line 9: shouldn't extends ExtendedBaseService be removed?

src/main/java/org/dcsa/core/events/service/VesselService.java
```diff
@@ -8,5 +8,11 @@
 
 public interface VesselService extends ExtendedBaseService<Vessel, UUID> {
```
pimg, line 9: shouldn't extends ExtendedBaseService be removed

pimg (changes requested):

src/main/java/org/dcsa/core/events/service/FacilityService.java
nt-as: No, UI Support (application supporting the UI) provides a facility controller that relies on the core features (filter + sorting).

src/main/java/org/dcsa/core/events/service/TimestampDefinitionService.java
nt-as: No, UI Support (application supporting the UI) provides a timestamp definition controller that relies on the core features (filter + sorting).

src/main/java/org/dcsa/core/events/service/VesselService.java
nt-as: No, UI Support (application supporting the UI) provides a vessel controller that relies on the core features (filter + sorting).

pimg (approved):

PR #226, opened 2022-02-21
DDT-753: move methods and models from SI

mp-assco:


src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
```diff
@@ -117,6 +159,41 @@
...
+    // TODO: we have a known bug here that needs to be addressed.
+    //  carrierBookingReference can differ from each cargoItem.
+    return shipmentRepository
+        .findByCarrierBookingReference(carrierBookingReference)
+        .switchIfEmpty(
+            Mono.error(
+                ConcreteRequestErrorMessageException.invalidParameter(
+                    "No shipment found with carrierBookingReference: " + carrierBookingReference)))
+        .flatMap(
+            x ->
```
pimg, line 176: use more appropriate name for example shipment

src/main/java/org/dcsa/core/events/service/ShipmentEquipmentService.java
```diff
@@ -10,4 +11,8 @@
 	Mono<List<ShipmentEquipmentTO>> createShipmentEquipment(UUID shipmentID, String shippingInstructionID, List<ShipmentEquipmentTO> shipmentEquipments);
 
 	Mono<List<ShipmentEquipmentTO>> findShipmentEquipmentByShipmentID(UUID shipmentID);
+
+	Mono<List<ShipmentEquipmentTO>> insertShipmentEquipmentTOs(List<ShipmentEquipmentTO> shipmentEquipments, ShippingInstructionTO shippingInstructionTO);
```
pimg, line 15: method name seems inappropriate. It has TO in the name whereas the others do not. Also the name insert feels ambiguis to create. "addShipmentEquipmentToShippingInstruction seems better

pimg (changes requested):

src/main/java/org/dcsa/core/events/service/impl/ShipmentEquipmentServiceImpl.java
mp-assco: Renamed to `shipment`.

src/main/java/org/dcsa/core/events/service/ShipmentEquipmentService.java
mp-assco: Renamed `insertShipmentEquipmentTOs` to `addShipmentEquipmentToShippingInstruction`.

pimg (approved):

PR #242, opened 2022-03-01
DDT-868: CargoItem and ShipmentEquipment is required on SI

nt-as:

### DIFF
--- a/src/main/java/org/dcsa/core/events/model/transferobjects/ShipmentEquipmentTO.java
+++ b/src/main/java/org/dcsa/core/events/model/transferobjects/ShipmentEquipmentTO.java
@@ -6,6 +6,7 @@ import lombok.NoArgsConstructor;
 import org.dcsa.core.events.model.AbstractShipmentEquipment;
 
 import javax.validation.Valid;
+import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.NotNull;
 import java.util.List;
 
@@ -19,6 +20,7 @@ public class ShipmentEquipmentTO extends AbstractShipmentEquipment {
 	EquipmentTO equipment;
 
 	@Valid
+	@NotEmpty
 	private List<CargoItemTO> cargoItems;
 
 	@Valid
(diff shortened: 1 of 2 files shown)


nt-as:
Relevant test changes are in https://github.com/dcsaorg/DCSA-EBL/pull/153

src/main/java/org/dcsa/core/events/model/transferobjects/ShipmentEquipmentTO.java
```diff
@@ -18,12 +19,14 @@
 	@NotNull(message = "Equipment is required.")
 	EquipmentTO equipment;
 
-	@Valid
-	private List<CargoItemTO> cargoItems;
+  @Valid
+  @NotNull
+  @Size(min = 1)
```
pimg, line 24: NotNull and Size can be simplified with @NotEmpty

src/main/java/org/dcsa/core/events/model/transferobjects/ShippingInstructionTO.java
```diff
@@ -26,6 +28,8 @@
 
   @JsonProperty("utilizedTransportEquipments")
   @Valid
+  @NotNull
+  @Size(min = 1)
```
pimg, line 32: same as above

pimg (commented):
some small comments

src/main/java/org/dcsa/core/events/model/transferobjects/ShipmentEquipmentTO.java
nt-as: Fixed

src/main/java/org/dcsa/core/events/model/transferobjects/ShippingInstructionTO.java
nt-as: Fixed

pimg (approved):

douradofrazer:
Should we add a message in the validation annotation? It helps for the returned error response.

PR #270, opened 2022-03-23
DDT-978: Added ValueAddedService to SI

mp-assco:


HenrikHL (approved):

nt-as (changes requested):
Henrik reported earlier today that `VAS` is being removed from the beta. Marking this a `Request changes` to avoid accidental approval/merge for now.

PR #271, opened 2022-03-24
DDT-981 [1/?]: Rename ShipmentEquipment to UtilizedTransportEquipment

nt-as:
This commit performs a rename of ShipmentEquipment (at a Java level)
to UtilizedTransportEquipment. The SQL rename will be handled in a
separate commit (when we have to inject the consignment item).


src/main/java/org/dcsa/core/events/repository/impl/CargoItemCustomRepositoryImpl.java
```diff
@@ -71,7 +71,7 @@
     //		join dcsa_im_v3_0.cargo_line_item cli on cli.cargo_item_id = ci.id
     //		where ci.shipment_equipment_id = :shipmentEquipmentID;
 
-    Objects.requireNonNull(shipmentEquipmentID, "ShipmentEquiment must not be null");
+    Objects.requireNonNull(utilizedTransportEquipmentID, "ShipmentEquiment must not be null");
```
HenrikHL, line 74:
```suggestion
    Objects.requireNonNull(utilizedTransportEquipmentID, "UtilizedTransportEquipment must not be null");
```

HenrikHL (approved):

PR #277, opened 2022-03-25
added ConsignmentItem entities

FinnAtAsseco:


src/main/java/org/dcsa/core/events/edocumentation/model/ConsignmentItem.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @NotNull
+  @Size(max = 10)
+  @Column("hs_code")
+  private String hsCode;
+
+  @NotNull
+  @Size(max = 100)
+  @Column("shipping_instruction_id")
+  private String shippingInstructionReference;
```
nt-as, line 33: Call it `ShippingInstructionID` for the DAO layer.  The column is called an ID and when we get entity versioning, the `ShippingInstruction` will get a `shippingInstructionReference` column (which is the external field rather than the internal one).

src/main/java/org/dcsa/core/events/edocumentation/model/transferobject/ConsignmentItemTO.java
```diff
@@ -0,0 +1,48 @@
...
+
+@Value
+@Builder
+public class ConsignmentItemTO {
+  public enum WeightUnit {
+    KGM, LBR
+  }
+  public enum VolumeUnit {
+    CBM, CFT
+  }
```
nt-as, line 23: These already exist (see how they used in `CargoItemTO` - possibly `AbstractCargoItem`)

src/main/java/org/dcsa/core/events/edocumentation/model/transferobject/ConsignmentItemTO.java
```diff
@@ -0,0 +1,48 @@
...
+  private VolumeUnit volumeUnit;
+
+  @NotBlank
+  private String descriptionOfGoods;
+
+  @Size(max = 10)
+  @JsonProperty("HSCode")
+  private String hsCode;
+
+  private List<CargoItemTO> cargoItems = Collections.emptyList();
```
nt-as, line 46: `@NonEmpty`

nt-as (approved):

PR #282, opened 2022-03-30
DDT-983: Update PUT SI

mp-assco:


src/test/java/org/dcsa/core/events/service/impl/ConsignmentItemServiceImplTest.java
```diff
@@ -0,0 +1,1063 @@
+//package org.dcsa.core.events.service.impl;
```
pimg, line 1: All these tests are commented out, why are they part of this PR in this form?

src/main/java/org/dcsa/core/events/edocumentation/service/impl/ConsignmentItemServiceImpl.java
```diff
@@ -99,6 +97,39 @@
...
+                    .flatMap(
+                        consignmentItem ->
+                            referenceRepository
+                                .deleteByConsignmentItemID(consignmentItem.getId())
+                                .flatMap(
+                                    x ->
+                                        consignmentItemRepository.deleteById(
+                                            consignmentItem.getId()))))
+        .collectList()
+        .flatMap(ignored -> Mono.empty());
```
pimg, line 130: I don't understand this part. Does it return a Mono.empty() or a list of consignment items?

src/test/java/org/dcsa/core/events/service/impl/ConsignmentItemServiceImplTest.java
mp-assco: Mistake. It was supposed to be in a separate PR for consignment tests.

src/main/java/org/dcsa/core/events/edocumentation/service/impl/ConsignmentItemServiceImpl.java
mp-assco: Empty. Made clearer now.

src/test/java/org/dcsa/core/events/service/impl/ConsignmentItemServiceImplTest.java
mp-assco: Removed

pimg (approved):

PR #283, opened 2022-04-01
DDT-983: Fix for CreateConsignmentItem

mp-assco:


pimg (changes requested):
Changes appear good, but tests are missing

mp-assco:
> Changes appear good, but tests are missing

Done

src/main/java/org/dcsa/core/events/edocumentation/service/impl/ConsignmentItemServiceImpl.java
```diff
@@ -119,13 +121,12 @@
...
-                                    x ->
-                                        consignmentItemRepository.deleteById(
-                                            consignmentItem.getId()))))
-        .then(Mono.empty());
+                            Mono.when(
+                                    referenceRepository.deleteByConsignmentItemID(
+                                        consignmentItem.getId()),
+                                    consignmentItemRepository.deleteById(consignmentItem.getId()))
+                                .thenReturn(consignmentItem)))
+        .then(Mono.just(shippingInstructionReference));
```
pimg, line 129: why do we return a reference which is no longer present? Other delete operations return void

pimg (changes requested):

src/main/java/org/dcsa/core/events/edocumentation/service/impl/ConsignmentItemServiceImpl.java
mp-assco: I wrote about this in the DCSA chat, but didn't really get a reply. The `assertNext` fails in the unit test when you return `void` for some reason and the only way I could find around it was to return a value. Would prefer just returning void, but don't know how to prevent errors in the tests.

pimg (approved):

PR #290, opened 2022-04-07
DDT-995 - find and replace cargo_item joins

FinnAtAsseco:


src/main/java/org/dcsa/core/events/repository/BookingRepository.java
```diff
@@ -76,9 +74,8 @@
 
   @Query(
       "SELECT DISTINCT b.* FROM shipping_instruction si "
-          + "JOIN cargo_item ci ON ci.shipping_instruction_id = si.id "
-          + "JOIN utilized_transport_equipment ute ON ute.id = ci.utilized_transport_equipment_id "
-          + "JOIN shipment s ON s.id = ute.shipment_id "
+          + "JOIN consignment_item ci ON ci.shipment_id = s.id "
+          + "JOIN shipment s ON s.id = ci.shipment_id "
           + "JOIN booking b ON b.id = s.booking_id "
```
nt-as, line 79: This looks wrong.  The first JOIN uses `s.id` but `s` is not defined at that time.

src/main/java/org/dcsa/core/events/repository/TransportRepository.java
```diff
@@ -22,11 +21,9 @@
...
-          + "JOIN utilized_transport_equipment ute"
-          + " ON ute.shipment_id = st.shipment_id"
-          + "JOIN cargo_item ci "
-          + " ON ci.utilized_transport_equipment_id = ute.id "
+          + " ON st.transport_id = t.id "
+          + "JOIN consignment_item ci "
+          + " ON st.shipment_id = ci.shipment_id "
           + "LEFT JOIN reference r "
           + "ON r.shipment_id = st.shipment_id "
           + "WHERE (ci.shipping_instruction_id = :shippingInstructionReference OR r.shipping_instruction_id = :shippingInstructionReference)")
```
nt-as, line 29: It looks like this query never worked (the original version reads `t.idJOIN` - the correction turns this into `t.id JOIN`), so this smells the code is unused.  Can it easily be removed instead? (to avoid churning dead code.  The document versioning might need to rework this query again soon)

src/main/java/org/dcsa/core/events/repository/BookingRepository.java
FinnAtAsseco: Fixed

src/main/java/org/dcsa/core/events/repository/TransportRepository.java
```diff
@@ -35,11 +32,9 @@
...
       "SELECT DISTINCT t.vessel_imo_number FROM transport t "
           + "JOIN shipment_transport st "
-          + " ON st.transport_id = t.id"
-          + "JOIN utilized_transport_equipment ute"
-          + " ON ute.shipment_id = st.shipment_id"
-          + "JOIN cargo_item ci "
-          + " ON ci.utilized_transport_equipment_id = ute.id "
+          + " ON st.transport_id = t.id "
+          + "JOIN consignment_item ci "
+          + " ON st.shipment_id = ci.shipment_id "
```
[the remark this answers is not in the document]
FinnAtAsseco: I will try to remove it and see where it is referenced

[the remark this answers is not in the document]
FinnAtAsseco: I'm thinking it's missing unit tests. This is called from EventSubscriptionServiceImpl in a special case where shipmentEvent.documentTypeCode is TRD

nt-as (approved):

PR #293, opened 2022-04-11
Ddt 960 update shipment events

pimg:
resolves: DDT-960

### DIFF
--- a/src/main/java/org/dcsa/core/events/model/ShipmentEvent.java
+++ b/src/main/java/org/dcsa/core/events/model/ShipmentEvent.java
@@ -1,6 +1,7 @@
 package org.dcsa.core.events.model;
 
 
+import com.fasterxml.jackson.annotation.JsonGetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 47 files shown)


nt-as (dismissed):

src/main/java/org/dcsa/core/events/model/ShipmentEvent.java
```diff
@@ -36,36 +37,21 @@
...
-
-    @JsonProperty("shipmentInformationTypeCode")
-    @Deprecated
-    public DocumentTypeCode getShipmentInformationTypeCode() {
-        if (documentTypeCode == DocumentTypeCode.BKG) {
-            return DocumentTypeCode.BOK;
-        }
-        return documentTypeCode;
-    }
-
```
nt-as, line 53: These are needed as backwards compat (TNT v2.2) as far as I remember.

nt-as: (ok with this as long as it is an explicit choice)

PR #294, opened 2022-04-11
DDT-1012: improvments based on json schema

Abdiiir:


src/main/java/org/dcsa/core/events/service/impl/DocumentPartyServiceImpl.java
```diff
@@ -154,17 +154,20 @@
...
-      });
+        .flatMap(
+            partyContactDetailsTOS -> {
+              if (partyContactDetailsTOS.isEmpty()) {
+                return Mono.error(
+                    ConcreteRequestErrorMessageException.notFound(
+                        "No contacts details were found for party"));
+              }
+              return Mono.just(partyContactDetailsTOS);
+            });
```
nt-as, line 170: This flatMap is redundant as we have a `@NonEmpty` check on that field AFAICT

douradofrazer (approved):

PR #298, opened 2022-04-12
DDT-1012: Improvments based on json schema

Abdiiir:


src/main/java/org/dcsa/core/events/edocumentation/service/impl/ConsignmentItemServiceImpl.java
```diff
@@ -63,7 +61,40 @@ public Mono<List<ConsignmentItemTO>> fetchConsignmentItemsTOByShippingInstructio
...
+                                              utilizedTransportEquipmentTOs -> {
+                                                Optional<UtilizedTransportEquipment> equipment =
+                                                    utilizedTransportEquipmentTOs.stream()
+                                                        .filter(
+                                                            x ->
+                                                                x.getId()
+                                                                    .equals(
+                                                                        cargoItem
+                                                                            .getUtilizedTransportEquipmentID()))
+                                                        .findFirst();
```
nt-as, line 81: This is basically a `HashMap`, please rewrite it as such.

(Map from ID to reference directly to simplify some of the following code).

pimg (approved):

PR #309, opened 2022-04-26
DDT-957 added query to retrieve active shipments and bookings

pimg:
Resolves the event-core part of DDT-957


nt-as (approved):

src/main/java/org/dcsa/core/events/repository/ShipmentRepository.java
```diff
@@ -16,6 +16,12 @@
 
   Mono<Shipment> findByCarrierBookingReference(String carrierBookingReference);
 
+  @Query(
+    "SELECT * FROM shipment S "
```
nt-as, line 20: Is that a capitial S?

PR #314, opened 2022-04-28
DDT-955: Update bookingStatusCheck

mp-assco:


src/main/java/org/dcsa/core/events/repository/BookingRepository.java
```diff
@@ -69,19 +69,27 @@ Mono<Booking> findByCarrierBookingRequestReferenceAndValidUntilIsNull(
 
   @Modifying
   @Query(
-    "UPDATE booking SET document_status = :documentStatus, updated_date_time = :updatedDateTime"
```
FinnAtAsseco, line 72: Is your indenting set to 4 instead of 2?

FinnAtAsseco (approved):

src/main/java/org/dcsa/core/events/repository/BookingRepository.java
mp-assco: no, it's set to 2. Don't know why it's auto-indenting line-breaks by 6 spaces...

PR #329, opened 2022-05-09
DDT-1039 added findTransportByShipmentID to transportService

pimg:
resolves: DDT-1039 (event_core part)


nt-as (approved):

src/main/java/org/dcsa/core/events/edocumentation/service/TransportService.java
```diff
@@ -7,4 +7,5 @@
 
 public interface TransportService {
     Flux<TransportTO> findByTransportID(UUID id);
+    Flux<TransportTO> findByShipmentID(UUID shipmentID);
```
nt-as, line 10: I assume this will be used in EBL?

PR #333, opened 2022-05-12
Rename Location in ShipmentLocation

mp-assco:


src/main/java/org/dcsa/core/events/edocumentation/model/transferobject/ShipmentLocationTO.java
```diff
@@ -14,7 +14,7 @@
 class ShipmentLocationTO {
 
   @NotNull(message = "Location is required.")
-  private LocationTO location;
+  private LocationTO locationTO;
```
nt-as, line 18: I suspect this will change the json name and cause us to no longer match the spec.  So this would need a `@JsonProperty("location")` to work.

mp-assco: 👍

pimg (approved):

PR #335, opened 2022-05-16
Query to use in unofficial endpoint for testing

mp-assco:


src/main/java/org/dcsa/core/events/repository/BookingRepository.java
```diff
@@ -98,4 +98,12 @@ Mono<Booking> findCarrierBookingReferenceAndValidUntilIsNull(
           + "JOIN booking b ON b.id = s.booking_id "
           + "WHERE si.shipping_instruction_reference = :shippingInstructionReference")
   Flux<Booking> findAllByShippingInstructionReference(String shippingInstructionReference);
+
+  @Query(
+      "SELECT b.* FROM dcsa_im_v3_0.transport_document td "
+          + "JOIN dcsa_im_v3_0.consignment_item ci ON td.shipping_instruction_id = ci.shipping_instruction_id "
```
pimg, line 104: dcsa_im_v3_0. this must be removed from all tables
use Java text blocks instead of string concat to make the query more readable

mp-assco: Done

mp-assco:
Found an existing way to do this instead.

